PR #18, opened 2016-03-14
Allow eq() and any() with UUID

K0zka:
Bug-Url: https://github.com/nhaarman/mockito-kotlin/issues/17
Signed-off-by: Laszlo Hornyak <laszlo.hornyak@gmail.com>


nhaarman:
Thanks.
Instead of explicitely fixing the issue for the `UUID` class, I have chosen to implement a more general fix.
See #19.

Issue #20, opened 2016-03-14
Allow createInstance() to use custom factory functions


nhaarman:
To avoid needing a library update for issues like #17, the `createInstance()` method should have some way of passing a custom initializer for certain classes.

PR #22, opened 2016-03-17
Add explicit capture with default value

neworld:
This PR will help for people like me, who have tons of final classes without an empty constructor. 

Also, it could solve #20

### DIFF
--- a/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/ArgumentCaptor.kt
+++ b/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/ArgumentCaptor.kt
@@ -29,6 +29,7 @@ import org.mockito.ArgumentCaptor
 
 inline fun <reified T : Any> argumentCaptor() = ArgumentCaptor.forClass(T::class.java)
 inline fun <reified T : Any> capture(captor: ArgumentCaptor<T>): T = captor.capture() ?: createInstance<T>()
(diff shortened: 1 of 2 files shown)


nhaarman:
Thank you for your input!
What do you think of #23 to solve this issue instead?

neworld:
#23 will fix this issue. But in my friendly opinion, it is not clean. The statics in tests adds complexity and  race-conditions. The problem is solved by registering instance creator in every class:
```
    @Test
    fun registerCreatorWithMock() {
        val constructorWithObject = ConstructorWithObject(mock())
        MockitoKotlin.registerInstanceCreator { constructorWithObject }
        val lambda: Function1<ConstructorWithObject, Unit> = mock()

        verify(lambda).invoke(any())
    }
```

But this adds more boilerplate. Also, the second example will fail, because you can't use mock() during verification:

```
    @Test
    fun registerCreatorWithMock() {
        MockitoKotlin.registerInstanceCreator { ConstructorWithObject(mock()) }
        val lambda: Function1<ConstructorWithObject, Unit> = mock()
        
        verify(lambda).invoke(any())
    }
```

nhaarman:
Good point.

neworld:
Because of global state. Since I don't know how to ensure initialize static before all tests, there are chances somebody will forget to init them, but tests still pass. Adding new tests, removing, altering them can change the order of tests and break that particular test.

neworld:
I realized this PR fixes a symptom, not [the real issue](https://github.com/nhaarman/mockito-kotlin/issues/24). #23 could work as a temporal solution. So, I close this PR.

PR #32, opened 2016-04-16
Updated kotlin and JUnit dependencies. Changed the JUnit because 1.10…

cesards:
- [ ] Can you back your code up with tests?

I had this problem, compiling my project: 
```
Error:Conflict with dependency 'org.jetbrains.kotlin:kotlin-stdlib'. Resolved versions for app (1.0.1-2) and test app (1.0.0) differ. See http://g.co/androidstudio/app-test-app-conflict for details.
Error:Conflict with dependency 'org.jetbrains.kotlin:kotlin-runtime'. Resolved versions for app (1.0.1-2) and test app (1.0.0) differ. See http://g.co/androidstudio/app-test-app-conflict for details.
```
Solved using the last kotlin dependency. Updated mockito dependency as well.

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'kotlin'
 
 buildscript {
-  ext.kotlin_version = '1.0.0'
+  ext.kotlin_version = '1.0.1-2'
 
   repositories {
     mavenCentral()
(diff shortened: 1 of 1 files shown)


nhaarman:
Thanks! However, this library does depend on the beta version of Mockito to make things work.

I will do a seperate update to update the dependency versions.

cesards:
👍

Issue #27, opened 2016-03-27
`uncheckedMock()` is not safe.


nhaarman:
This issue investigates the cause of #24.

The `uncheckedMock` function bypasses the internal Mockito state checks to mock objects during verification. No interaction with these mocks is allowed, by the internals of Mockito.

However, when creating an unchecked mock of a `Throwable` class which takes a `Throwable` itself as a parameter, the `Throwable` class calls `toString()` on the unchecked mock, causing the exception below.

```kotlin
class SomeThrowable : Throwable {

    constructor(cause: Throwable) : super(cause) {
    }
}

class SomeTest {

    @Test
    fun test() {
        any<SomeThrowable>()
    }
}
```

Stack trace:

```
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at kotlin.reflect.jvm.internal.FunctionCaller$Constructor.call(FunctionCaller.kt:63)
	at kotlin.reflect.jvm.internal.KCallableImpl$DefaultImpls.call(KCallableImpl.kt:65)
	at kotlin.reflect.jvm.internal.KFunctionImpl.call(KFunctionImpl.kt:30)
	at kotlin.reflect.jvm.internal.KCallableImpl$DefaultImpls.callBy(KCallableImpl.kt:101)
	at kotlin.reflect.jvm.internal.KFunctionImpl.callBy(KFunctionImpl.kt:30)
	at com.nhaarman.mockito_kotlin.CreateInstanceKt.newInstance(CreateInstance.kt:136)
	at com.nhaarman.mockito_kotlin.CreateInstanceKt.createInstance(CreateInstance.kt:60)
	at tmp.SomeTest.test(SomeTest.kt:45)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #37, opened 2016-05-15
Fixed ClassCastException thrown when using with PowerMock

jskierbi:
There is ClassCastException when using this lib with PowerMock. I added class check before casting inside CreateInstance.kt, which prevents the exception. No PowerMock dependency introduced.

**Problem description**

Exception:
```
java.lang.ClassCastException: com.amazon.identity.auth.device.shared.APIListener$$EnhancerByMockitoWithCGLIB$$6966922d cannot be cast to org.mockito.internal.creation.bytebuddy.MockMethodInterceptor$MockAccess

	at com.nhaarman.mockito_kotlin.CreateInstanceKt.uncheckedMock(CreateInstance.kt:165)
	at com.nhaarman.mockito_kotlin.CreateInstanceKt.createInstance(CreateInstance.kt:55)
        ...
```

Code that caused the exception
```kotlin
@RunWith(PowerMockRunner::class)
class MockitoKotlinPowermockTest {
    @Test fun anyTest() {
        any<APIListener>() // Class cast exception inside
    }
}
```

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -22,9 +22,15 @@ dependencies {
   compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
   compile "org.mockito:mockito-core:2.0.52-beta"
 
+
   /* Tests */
   testCompile "junit:junit:4.12"
   testCompile "com.nhaarman:expect.kt:0.5.1"
+  testCompile "org.jetbrains.kotlin:kotlin-test:$kotlin_version"
+
+  def powermockVersion = "1.6.5"
+  testCompile "org.powermock:powermock-module-junit4:${powermockVersion}"
+  testCompile "org.powermock:powermock-api-mockito2:${powermockVersion}"
 }
 
 publishing {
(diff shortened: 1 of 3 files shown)


marcinbak:
+1

nhaarman:
Thanks!
I'm not at all familiar with PowerMock though, and I'd like a regression test to make sure things won't break in the future.

Could you perhaps create a test that fails without this modification? I don't mind if you have to add additional dependencies, I'm willing to clean up after you :)

jskierbi:
I'll write some tests, np. Just found out that this workaround misbehaves a bit - after mocking an object real method is called, which is obviously not good. I'll dig up a bit following days and try to push update to this PR both with tests and fix.

nhaarman:
What is the current status on this?

nhaarman:
The ClassCastException should be fixed by 85bb5e4c7941bf6e34df6dfa17e1eafbb0ebe7d7. This may however cause #27 to reappear for PowerMock users, I'm not sure of that.

PR #40, opened 2016-06-20
Use nulls not instances

Sarah-Gallitz:
This successfully satisfies the mockito library but greatly simplifies the code. Instance creators are no longer needed.

- [x]  Can you back your code up with tests?
- [x]  Keep your commits clean: squash your commits if necessary.
- [x]  Make sure you're targeting the dev branch with the PR.


nhaarman:
IIRC I have read somewhere that this is undocumented behavior and subject to change.
Though this would really make life a lot easier, I am not entirely sure if we can depend on this.

I will look into it.

nhaarman:
See https://youtrack.jetbrains.com/issue/KT-8135.

Sarah-Gallitz:
Fair enough. It seems they acknowledge something like this is necessary for testing, but that was in 2015. 
Should I close the PR until an official stance is taken?

nhaarman:
Let's close this for now, the `mock-maker-inline` feature eases things as well.

PR #49, opened 2016-07-21
Makes `mock` return non-null types.

Zoltu:

 - [ ] Can you back your code up with tests?
 - [ ] Keep your commits clean: [squash your commits if necessary](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
 - [ ] Make sure you're targeting the `dev` branch with the PR.

Since `Mockito.mock` is a Java function it returns a Kotlin platform type.  By adding an explicit return type to the wrapper, the kotlin version of `mock` will now return a non-platform type.  In this case, we know that `Mockito.mock` doesn't return null so we can safely make the wrapper return a non-null type.  We add the `!!` to make it clear to the reader our intent is to strip the platform type at this point, and also make the Kotlin compiler add an explicit null-check "just in case" before returning from the kotlin `mock` method.


Zoltu:
Switching this to PR to `dev` per instructions.

PR #50, opened 2016-07-21
Makes `mock` return non-null types.

Zoltu:

Since `Mockito.mock` is a Java function it returns a Kotlin platform type.  By adding an explicit return type to the wrapper, the kotlin version of `mock` will now return a non-platform type.  In this case, we know that `Mockito.mock` doesn't return null so we can safely make the wrapper return a non-null type.  We add the `!!` to make it clear to the reader our intent is to strip the platform type at this point, and also make the Kotlin compiler add an explicit null-check "just in case" before returning from the kotlin `mock` method.


Zoltu:
Note to maintainer: Recommend making `dev` the default branch in GitHub (https://help.github.com/articles/setting-the-default-branch/) so that people submitting PRs automatically "do the right thing".

Zoltu:
No tests since this is really just a quality-of-life tooling related improvement.  If someone can think of a way to actually test this let me know, but since `Mockito.mock` can't return a null I'm not sure how to test this change in a meaningful way.

nhaarman:
Thanks!

PR #55, opened 2016-08-09
Upgrade Mockito to 2.0.99-beta

sghill:
Recent versions of Mockito break with mockito-kotlin because of changes in mockito/mockito#404:

* org.mockito.internal.util.MockUtil is now a static utility class
* org.mockito.Mockito#stub has been removed in favor of #when

This change updates mockito-kotlin to not instantiate MockUtil and aliases `com.nhaarman.mockito_kotlin.stub` to `com.nhaarman.mockito_kotlin.whenever`, which was suggested by the docs. 

I think it may be better to remove the stub method all together, since Mockito has chosen to do that, but figured this was a safer starting point. Happy to update the PR if that's the better route.

---

 - [x] Can you back your code up with tests? *Underlying Mockito API changes; all existing tests pass*

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -20,7 +20,7 @@ repositories {
 dependencies {
   compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
-  compile "org.mockito:mockito-core:2.0.52-beta"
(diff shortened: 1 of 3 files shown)


nhaarman:
Thanks!

Merely replacing the `stub` call by `whenever` will still break any client's code since their return types differ.

Since this library is essentially a wrapper for the original Mockito library, I think we should indeed remove the `stub` method and keep the api one-on-one.

sghill:
Sounds good to me -- PR has been updated to remove `stub`

PR #78, opened 2016-10-02
Update build.gradle

sdoward:
I don't really have time to pull the repo and run the tests so instead of opening an issue I just bumped the kotlin version via the web UI as a friendly reminder.

Thanks for the good work


sdoward:
I just saw somebody else had the same idea so I am closing the PR

PR #86, opened 2016-10-04
enable mocking final classes by default

oshai:
see: http://hadihariri.com/2016/10/04/Mocking-Kotlin-With-Mockito/
I placed it in src and not test because the lib is used with test
dependency itself. if it will be in test other projects will not be
getting the resource.


nhaarman:
Not choosing to do this at the moment, because the feature is incubating and opt-in in Mockito itself. Once it's stable enough and no real issues arise from it, we might include this too. May it be that Mockito handles this for us though.

PR #88, opened 2016-10-11
#37 - Fix ClassCastException introduced by Mockito's mock-maker-inline

desilvai:
I saw another pull request that has been in limbo for a while.  I experienced the same issue in Mockito's new mock-maker-inline.  While this feature for Mockito is incubating, it highlighted a flawed assumption in the code for mockito-kotlin: that all created mocks subclass MockAccess.  This doesn't hold when different mocking strategies (plugins) are used (hence why it fails for mock-maker-inline and not for open classes with open functions).  This is likely why @jskierbi observed a similar issue with PowerMock.  He mentioned the proposed fix (essentially the same as the one I've included) didn't quite work with PowerMock.  So far, I have not observed anything to indicate that is true for the mock-maker-inline feature.  I included a similar testing approach as well to ensure that we would get an error if the original method was ever called.

While I wouldn't normally suggest fixing stuff for an incubating feature, this seems quite important since it allows us to mock Kotlin classes and stub methods without declaring them open.  We can still get the benefits of some of the Kotlin protections while still creating robust tests with mocks.  It also reduces the burden for anyone starting with mocking on an already large project.

Because of the way we have to enable this feature (by creating a resource that triggers a different mocking strategy applied to all tests--see http://hadihariri.com/2016/10/04/Mocking-Kotlin-With-Mockito/), I placed my tests in a different source folder.  This way, existing tests that are expected to fail due to a class/method being closed will continue to do so.


nhaarman:
Thanks! I have made some modifications to match the code style and added some additional checks. 
See 85bb5e4c7941bf6e34df6dfa17e1eafbb0ebe7d7.

PR #132, opened 2016-12-06
Allows fluent mock definition (calling X returns Y) way

passsy:
Allows fluent mock definition in a `"calling X returns/throws Y"` way

Example1:
```kotlin
        val mock: Thread = mock {
            calling { toString() } returns "Hello Kotlin"
            calling { run() } throws RuntimeException("boom")
        }
```

Example2:
```kotlin
        val mock: Thread = mock {
            calling(it.toString()) returns "Hello Kotlin"
            calling(it.run()) throws RuntimeException("boom")
        }
```
---
If this change doesn't get in or while this PR is open the following extension functions can be used:

```kotlin
fun <R, T> KStubbing<T>.calling(methodCall: T.() -> R): OngoingStubbing<R> = on(methodCall)
fun <R> calling(methodCall: R): OngoingStubbing<R> = Mockito.`when`(methodCall)

infix fun <T> OngoingStubbing<T>.returns(t: T): OngoingStubbing<T> = thenReturn(t)
infix fun <T> OngoingStubbing<T>.throws(t: Throwable): OngoingStubbing<T> = doThrow(t)
```

### DIFF
--- a/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/Mockito.kt
+++ b/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/Mockito.kt
@@ -95,6 +95,7 @@ inline fun <reified T : Any> mock(stubbing: KStubbing<T>.(T) -> Unit): T {
 
 class KStubbing<out T>(private val mock: T) {
     fun <R> on(methodCall: R) = Mockito.`when`(methodCall)
+    fun <R> calling(methodCall: R) = Mockito.`when`(methodCall)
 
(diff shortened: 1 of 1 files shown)


nhaarman:
Thanks! This looks pretty good.
I'm a bit weary to adding _another_ way of doing the same thing, to avoid cluttering the public API.

I will leave this open for the moment.

passsy:
Next step, deprecate `on` and `thenReturn` in favour of `calling` and `returns` 😉

nhaarman:
Yep, but that would result in a `2.x` release ;)

nhaarman:
Closing this as I don't see it becoming part of the API any time soon.

passsy:
I'm still a big fan of this syntax and add it to all projects! 😬

PR #158, opened 2017-03-30
Added stub extension function

pbochenski:
with this method one can define mock for class

`val someMock = mock<Object>()`

and then define stubs in test

`someMock.stub {  
    on { something } doReturn somethingElse  
}`

It is useful when some tests reuse mock object but with different stubbing


nhaarman:
Nice idea! Could you perhaps add some regression tests, like https://github.com/nhaarman/mockito-kotlin/blob/master/mockito-kotlin/src/test/kotlin/test/MockitoTest.kt#L375 ?

pbochenski:
please let me know if this is enough, or should I add all variants of stubbing to this.

nhaarman:
Spot on, thanks!

PR #171, opened 2017-05-20
Added support for coroutines

Dreckr:
This change bumps Kotlin version to 1.1.1 and adds support for coroutines.

There was a fix in Mockito that made it work with Kotlin coroutines (https://github.com/mockito/mockito/pull/1032), so its version has also been bumped.

A new test case was added to ensure that suspending functions are well handled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,7 @@ language: java
 matrix:
   include:
   - jdk: oraclejdk7
-    env: TERM=dumb MOCK_MAKER=mock-maker-inline KOTLIN_VERSION=1.0.7
-  - jdk: oraclejdk7
     env: TERM=dumb MOCK_MAKER=mock-maker-inline KOTLIN_VERSION=1.1.1
-  - jdk: oraclejdk8
-    env: TERM=dumb KOTLIN_VERSION=1.0.7
   - jdk: oraclejdk8
     env: TERM=dumb KOTLIN_VERSION=1.1.1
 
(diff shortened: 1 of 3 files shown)


Dreckr:
This breaks compatibility with Kotlin 1.0.7, as coroutines were added on 1.1

nhaarman:
I don't think this is really necessary, apart from updating the Mockito version.
You're basically only including the `kotlinx:kotlinx-coroutines-core` as an extra dependency which is something the users can do themselves, right?

For current users, explicitly depending on Mockito `2.8.29` and enabling coroutines should do the trick as well.

Dreckr:
Fair enough. So I'll just leave this as a reference for anyone having the same problems that I did:

If you are getting exceptions like this:
```
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Invalid use of argument matchers!
2 matchers expected, 1 recorded:
-> at com.nhaarman.mockito_kotlin.MockitoKt.eq(Mockito.kt:114)

This exception may occur if matchers are combined with raw values:
    //incorrect:
    someMethod(anyObject(), "raw String");
When using matchers, all arguments have to be provided by matchers.
For example:
    //correct:
    someMethod(anyObject(), eq("String by matcher"));

For more info see javadoc for Matchers class.
```
and your matchers are all correct but you are mocking the call of a suspending function, you'll need to force your Mockito to a newer version because coroutines were not supported.

Support to coroutines was added to Mockito on `2.8.27` which is only available on https://bintray.com/mockito/maven/mockito-development for now, so you'll need to add it as a maven repository of your project. Mockito's next 'Notable' release (currently at version `2.8.9`) should include this changes, so you won't need to add this repository.

If you have any doubts, take a look at the changes made on this PR.

Issue #163, opened 2017-04-14
Usage with kotlin 1.1.1


dpoetzsch:
Hi,

we use kotlin 1.1.1. When trying to add mockito-kotlin to our dependencies the gradle build fails with the following error:

```
WARNING: Conflict with dependency 'org.jetbrains.kotlin:kotlin-stdlib'. Resolved versions for app (1.1.1) and test app (1.0.7) differ. See http://g.co/androidstudio/app-test-app-conflict for details.

....

Conflict with dependency 'org.jetbrains.kotlin:kotlin-stdlib'. Resolved versions for app (1.1.1) and test app (1.0.7) differ. See http://g.co/androidstudio/app-test-app-conflict for details.
```

Is it possible to use mockito-kotlin with kotlin 1.1.1?

nhaarman:
See the [FAQ](https://github.com/nhaarman/mockito-kotlin/wiki/FAQ#dependencyconflict).

dpoetzsch:
Sorry overlooked that one. Thanks

Issue #162, opened 2017-04-07
Kotlin 1.1 Migration


kiwiandroiddev:
Just wondering if bumping to Kotlin 1.1 is something in the pipeline?

Thanks for your work on this library - we're making heavy use of it on an Android project.

Edit: just saw this commit: https://github.com/nhaarman/mockito-kotlin/commit/a52ae865d89ab50f804fc4938f44bdcaeb4e5754
Disregard :)

nhaarman:
The library is built with Kotlin 1.0.x, but works with 1.1.x just fine. The main reason for not bumping the library's dependency to 1.1.x is to make sure no 1.1.x features are used that will break users who are still on 1.0.x.

If you get an error which mentions a conflict between 1.0.x and 1.1.0, see [this page](https://github.com/nhaarman/mockito-kotlin/wiki/FAQ#dependencyconflict).

Issue #156, opened 2017-03-01
Kotlin 1.1.0 compatibility


gray419:
Great library, excited to see Kotlin 1.1 compatibility

mg6maciej:
It is 1.1 compatible. The only thing you need to do to use it, if you don't depend on kotlin-reflect in other ways than as a transitive dependency in mockito-kotlin, is to add:
```
    testCompile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
```
and / or on Android:
```
    androidTestCompile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
```

nhaarman:
When I test a project that uses `1.1.0` and depends on Mockito-Kotlin with Kotlin `1.0.x`, I only get this warning:

```
w: Runtime JAR files in the classpath should have the same version. These files were found in the classpath:
    (...)/kotlin-reflect-1.0.6.jar (version 1.0)
    (...)/kotlin-stdlib-1.1.0.jar (version 1.1)
w: Consider providing an explicit dependency on kotlin-reflect 1.1 to prevent strange errors
```

I am not using instrumentation tests though. As I understand correctly, your build fails if you don't include those lines?

mg6maciej:
Yes, even `./gradlew test` gives this:
```
:app:prepareDebugUnitTestDependencies
Conflict with dependency 'org.jetbrains.kotlin:kotlin-stdlib' in project ':app'. Resolved versions for app (1.1.0) and test app (1.0.6) differ. See http://g.co/androidstudio/app-test-app-conflict for details.
:app:prepareDebugUnitTestDependencies FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:prepareDebugUnitTestDependencies'.
> Dependency Error. See console for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

which is kind of weird, as the tests are run on the computer and not emulator / device.
Same error when running `./gradlew connectedAndroidTest`.

nhaarman:
Ok, so I see the issue here.

In my own project, I only declare the following dependencies:

```
compile kotlinStdLib
testCompile mockitoKotlin
```

The `stdLib` dependency uses `1.1.0`, where `mockitoKotlin` pulls in `1.0.x`.
[...]

PR #178, opened 2017-06-03
Create module to release a kt1.1 artifact

nhaarman:
This should generate another artifact that includes the `1.1.x` version of Kotlin, and fix issues like #156, #162 and #163.


sdeleuze:
@nhaarman Why not just avoiding to expose `kotlin-stdlib` and `kotlin-reflect` dependencies in the Maven dependencies by using a scope like `compileOnly` and `compileTestOnly`,

That would avoid creating specific artifact for Kotlin 1.1, 1.2, 1.3 ...

nhaarman:
While the `stdlib` will probably always be included by the client, `reflect` might not.

sdeleuze:
I agree and I don't say the approach I propose is perfect but I think this is the one adapted by very well known project like Jackson Kotlin and I am really not a fan of seeing a project that propose `mockito-kotlin-kt1.1`, `mockito-kotlin-kt1.2` and `mockito-kotlin-kt1.3` artifact in less than a year.

I will ask on Kotlin Slack if somebody has a better solution to this issue.

sdeleuze:
I have created https://youtrack.jetbrains.com/issue/KT-18398 to try to find a better outcome than Kotlin version specific artifacts.

PR #182, opened 2017-07-02
Add dsl stubbing to spied classes.

Tapchicoma:
Now it is possible to describe spied classes mock behaviour in dsl form as
it was already possible with common mocks.

For example:

```
val spiedDate = spy(Date(123L)) {
    on { time } doReturn 12L
}
```

Signed-off-by: Yahor Berdnikau <egorr.berd@gmail.com>


nhaarman:
I have rebased your PR onto the `2.x` branch. See #221.

PR #188, opened 2017-07-11
Use non-deprecated version of constructor to set RETURNS_DEEP_STUBS

inktomi:
Very small update to test code to use the non-deprecated way of setting RETURNS_DEEP_STUBS. This is helpful because people are likely to use the tests as a sample of how best to do things like this.


nhaarman:
Included in a25512a25f2c771d0028ac5646d2b787a904ed0a

PR #197, opened 2017-08-31
Update Mockito to 2.9.0

hoombar:

 - [x] Can you back your code up with tests? - all tests still pass


nhaarman:
Thanks! Since 2.10.0 has been release in the mean while, I'll update to that directly.

Issue #205, opened 2017-09-28
Coroutine/suspend function mocking support


dave08:
If you use non-inline lambdas that are not declared as suspend, the runBlocking { } around the test function doesn't help, rather we need to `mock { on { runBlocking { suspendFun() } } }` for each mock...

nhaarman:
Could you post a more elaborate example?

dave08:
```kotlin
interface AccountRepository {
    suspend fun getAccountId(deviceNo: String): Int
}

class SUT(val repo: AccountRepository) {
   fun something() {
        launch(CommonPool) { println(repo.getAccountId("0000")) }
  }
}

// Test function
val repoMock:AccountRepo = mock { onGeneric { runBlocking { getAccountId(any()) } } doReturn 3300 }

val sut = SUT(repoMock)
sut.something()

// Do some assertions on mock
```

dave08:
If the mock and onGeneric lambdas were either suspend functions or inline, then I wouldn't need to put runBlocking inside the onGeneric for every function, I could just surround the whole mock {} with runBlocking, it would be much cleaner.

540Grunkspin:
Isn't this gonna be sort of hard to implement if you want to have the library compatible with pre 1.1 kotlin? Maybe soon Gradle magic?

540Grunkspin:
Or maybe it's enough to just make them inline.

dave08:
It might be enough to make them into extension functions that are inline,
I'm just not sure about nested dsl working that way...

בתאריך 2 באוק׳ 2017 7:55 PM,‏ "540Grunkspin" <notifications@github.com> כתב:

540Grunkspin:
I created a pull request for a suggested fix. Seems like all the previous tests are passing but still waiting for the CI to run.

Not sure if this is exactly what you wanted. There is an example test in there as well.

dave08:
Thanks, I'm on vacation for a week, but when I get back I'll try it out!
I'm sure it'll help lots of people, if coroutines pick up as much as Kotlin
did...

בתאריך 3 באוק׳ 2017 6:46 PM,‏ "540Grunkspin" <notifications@github.com> כתב:

PaulWoitaschek:
Any news on this? It's the same with `given`

```
    given { runBlocking { mockLoginManager.refreshToken() } }
        .willReturn {
          TokenFactory.create(expired = false)
        }
```

Creates quite some noise whenever there is a suspending function.

[...]

PR #206, opened 2017-10-03
Change on to be inline in order to support coroutines

540Grunkspin:
Suggested fix for #205 

Seems like all the previous tests work. Had to add the tests in the kt1.1 artifact in order to test the coroutines.

### DIFF
--- a/mockito-kotlin-kt1.1/build.gradle
+++ b/mockito-kotlin-kt1.1/build.gradle
@@ -27,6 +27,17 @@ dependencies {
     compile project(':mockito-kotlin')
     compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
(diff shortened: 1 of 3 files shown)


dave08:
Thanks! What about onGeneric?

540Grunkspin:
True. I can try adding inline to that I've too. Should work the same way

dave08:
Thanks @540Grunkspin !

540Grunkspin:
I found some problems with this method. Looks like if the return is a deferred and you await that you actually get back null every time. I will try to investigate this my self but all help would be appreciated.

540Grunkspin:
So await doesn't work basically

nhaarman:
@540Grunkspin 

> I found some problems with this method. Looks like if the return is a deferred and you await that you actually get back null every time. I will try to investigate this my self but all help would be appreciated.

Do you have an example test that shows this?

nhaarman:
I'm looking into a specialized version for coroutine stubbing. Please continue discussion in  #205

PR #214, opened 2017-11-29
Support Kotlin 1.2.0 and set default Kotlin version to 1.1.50

takuji31:
Added Kotlin 1.2.0 support.

- Use kotlin.reflect.full.starProjectedType instead of kotlin.reflect.defaultType
  - extensions in `jvm.reflect` package was removed in Kotlin 1.2.0
- Set default Kotlin version to 1.1.50
  - `kotlin.reflect.full` package was added in Kotlin 1.1.0
  - **[important]** Applying this change makes it impossible to support Kotlin 1.0.x.
- https://kotlinlang.org/docs/reference/whatsnew11.html#kotlinreflectfull
- https://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages

We want to use Kotlin 1.2.0 as soon as possible. Please check this 🙇

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,13 @@ language: java
 matrix:
   include:
   - jdk: oraclejdk8
-    env: TERM=dumb MOCK_MAKER=mock-maker-inline KOTLIN_VERSION=1.0.7
-  - jdk: oraclejdk8
     env: TERM=dumb MOCK_MAKER=mock-maker-inline KOTLIN_VERSION=1.1.50
   - jdk: oraclejdk8
-    env: TERM=dumb KOTLIN_VERSION=1.0.7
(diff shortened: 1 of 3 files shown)


nhaarman:
> [important] Applying this change makes it impossible to support Kotlin 1.0.x.

Well this is kind of a deal breaker. Does the current version of Mockito-Kotlin (`1.5.0`) not work for you with Kotlin `1.2.0`?

nhaarman:
Version `2.x` will remove the dependency on `kotlin-reflect`. Please try out `2.0.0-alpha01` to see if it works for you: https://github.com/nhaarman/mockito-kotlin/releases/tag/2.0.0-alpha01

takuji31:
I'll try using the new version soon. thanks 😄

takuji31:
New version works fine. 👍

PR #219, opened 2017-12-03
Allow nullable return types for `onGeneric`

### LINKED ISSUE #208: onGeneric does not work with nullable return type
The following code will not work due to the nullable return type:

```kotlin
val result: Iterable<String> = mock {
    onGeneric { firstOrNull() } doReturn mockedResponse
}
```

error thrown during compilation:
```
Kotlin: Type parameter bound for R in inline fun 

<reified R : Any> onGeneric(noinline methodCall: Iterable<String>.() -> R): OngoingStubbing<R>

is not satisfied: inferred type String? is not a subtype of Any
```

nhaarman:
Fixes #208


walkerdb:
🎉

PR #226, opened 2018-01-06
extension function for spies

dpreussler:
spy should make use of extensions functions, as they are based on existing objects, so no need to use same syntax as mock :)

it reads much nicer:

```kotlin
      val dateSpy = Date(0).spy()
```
or
```kotlin
      val dateSpy = Date(0).spy {
            on { time } doReturn timeVal
        }
```

had to rename to *spied* as would have the same jvm syntax as the latest merge to spy syntax

```kotlin
      val dateSpy = Date(0).spied()
```


nhaarman:
Using the parameter (`spy(instance)`) expresses the fact that the instance is wrapped to be spied, whereas `instance.spy()` might indicate that the instance is altered.
Since the `spy(instance) { }` syntax is the same as `mock() { } `) which people know I don't think this really adds any value to the library.

PR #228, opened 2018-01-11
Coroutines testing

### LINKED ISSUE #205: Coroutine/suspend function mocking support

bohsen:
Fixes #205 and shows how to test different uses of coroutines:
1. Using a coroutine with a suspend function and how to test this
2. Using a coroutine with a regular function and how to test this.


nhaarman:
Awesome, thanks! I'll look into this soon.

bohsen:
Your welcome. Great library.

Btw. Your build for kotling 1.0.7 will continue to fails because [support for coroutines was added in kotlin 1.1](https://kotlinlang.org/docs/reference/coroutines.html)

Issue #229, opened 2018-01-12
Consider updating build to use at least kotlin 1.1.0


bohsen:
This would add:
1. Support for suspending functions
2. Support for nullable types in argumentCaptor function (Maybe elsewhere too?)
`val captor = argumentCaptor<String?>()`

You could change ArgumentCaptor to this:

`inline fun <reified T : Any?> argumentCaptor(): KArgumentCaptor<T> = KArgumentCaptor(ArgumentCaptor.forClass(T::class.java), T::class)`

and maybe get rid of nullableArgumentcapture function, which is difficult to use unless you now how to do it.

nhaarman:
This is indeed something I want to do. I still want to make sure Kotlin 1.0.7 is supported, so I'll have to think of some CI thingy to make that work.

bohsen:
👍 I just noticed that version 2.0 mght be upcoming. That's always a good time to for som clean up.

PR #242, opened 2018-02-18
Simplified argument captor API

igorwojda:
This will simplify ArgumentCaptor code

```
        //old (still working)
        argumentCaptor<String>().apply {
            verify(mailRepository).sendEmail(capture())
            firstValue shouldEqual "A"
        }

        nullableArgumentCaptor<String>().apply {
            firstValue shouldEqual null
        }

        // new simplified overload usage
       argumentCaptor<String> {
            verify(mailRepository).sendEmail(capture())
            firstValue shouldEqual "A"
        }

        nullableArgumentCaptor<String> {
            firstValue shouldEqual null
        }
```

I needed to add another overload since older version of Kotlin does not support default values for functional parameters

### DIFF
--- a/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/ArgumentCaptor.kt
+++ b/mockito-kotlin/src/main/kotlin/com/nhaarman/mockito_kotlin/ArgumentCaptor.kt
@@ -29,8 +29,10 @@ import com.nhaarman.mockito_kotlin.createinstance.createInstance
 import org.mockito.ArgumentCaptor
 import kotlin.reflect.KClass
 
-inline fun <reified T : Any> argumentCaptor(param: KArgumentCaptor<T>.() -> Unit = {}): KArgumentCaptor<T> = KArgumentCaptor(ArgumentCaptor.forClass(T::class.java), T::class)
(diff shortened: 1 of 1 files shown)


igorwojda:
I also wonder why do we need `nullableArgumentCaptor` function? seems like specifying ` T : Any?` as type parameter bound does the trick:

```
inline fun <reified T : Any?> argumentCaptor(): KArgumentCaptor<T> = KArgumentCaptor(ArgumentCaptor.forClass(T::class.java), T::class)

inline fun <reified T : Any?> argumentCaptor(param: KArgumentCaptor<T>.() -> Unit): KArgumentCaptor<T> =
        KArgumentCaptor(ArgumentCaptor.forClass(T::class.java), T::class)
```

Now `nullableArgumentCaptor` seems to be redundant unless I am missing something

bohsen:
Same as #229

nhaarman:
The `nullableArgumentCaptor` is introduced to work with `allValues` and nullability: https://github.com/nhaarman/mockito-kotlin/pull/100

nhaarman:
You were not doing anything with the `param` parameter, so the code was never executed. I have fixed this, rebased onto `2.x` and opened #254 for this.

PR #257, opened 2018-05-31
Updated kotlin version to 1.2.41

CDehning:
- Updates kotlin to 1.2.41
- Makes the kotlin version variable available across all modules
- Uses gradle project dependencies inside test module, so that you can just run ./gradlew test without manually compiling first


nhaarman:
> Uses gradle project dependencies inside test module, so that you can just run ./gradlew test without manually compiling first

This is done intentionally for running tests against multiple Kotlin versions.
The Kotlin version is updated to 1.2.50 in #265

PR #273, opened 2018-08-09
Fix gradle reference in readme file

vitorprado:
Dependency for 1.x versions was missing in readme file


nhaarman:
Since 2.x is RC we'll keep this as it is. The `1.x` branch contains the proper references for 1.x.

PR #275, opened 2018-08-13
testCompile got replaced by testImplementation

ChristianNavolskyi:


nhaarman:
This library is not just for Android usage, and Android Studio will give warnings any when `testCompile` is used.

PR #278, opened 2018-08-22
Inline `whenever` to let Mockito's UnfinishedStubbing messages work

trav:
When Mockito detects unfinished stubbing, it will spit out an error containing:
```
Unfinished stubbing detected here:
-> at com.nhaarman.mockitokotlin2.OngoingStubbingKt.whenever(OngoingStubbing.kt:42)
```

This is not useful in helping users track down unfinished stubbing.  Instead, we inline `whenever`
to make it read as:
```
Unfinished stubbing detected here:
-> at com.MyTest.testMyThing(MyTest.kt:785)
```


nhaarman:
Nice one, thanks!

PR #280, opened 2018-08-26
provide InvocationOnMock instance for BDDMyOngoingStubbing#willAnswer

unlimitedsola:
This PR adds `InvocationOnMock` argument to `willAnswer`'s lambda


nhaarman:
This is a possible breaking change. What use cases is this necessary for? Also, can you provide a test that proves this works?

unlimitedsola:
My use case is that I can get arguments from `InvocationOnMock` when the generated stub got invoked.

nhaarman:
Could you include a test that demonstrates its use case?

PR #281, opened 2018-08-26
add argThat(ArgumentMatcher<T>) function

sothawo:
this PR adds the `Mockito.argThat(ArgumentMatcher<T>)` call to this library


nhaarman:
Thanks! Can you provide a test that shows a use case where this is necessary and the existing `argThat` function doesn't suffice?

sothawo:
I added some tests.
I came upon this when answering a question on SO (https://stackoverflow.com/questions/52005773/match-varargs-in-mockito-2-stubbing/52016020) where the need came up to have a vararg matcher tha can answer questions like _does any of the args apply to some predicate_ or _do all have the same predicate_. For this one would need to pass in an ArgumentMatcher instance that keeps state.

nhaarman:
Thanks! I'll make some minor modifications and will merge it in.

PR #289, opened 2018-09-12
Update README.md: Configuration 'compile' is obsolete and has been replaced 'implementation'

rkonno:
The testCompile configuration inherited from the Java plugin are still available but are deprecated. 
[The Java Library plugin configurations](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph)

It is a minor modification... 
Thank you.


nhaarman:
Thanks!

Issue #233, opened 2018-01-20
argumentCaptor for multiple arguments/types


dsebastien:
Is it or would it be possible to do something like this:

```
argumentsCaptor<Foo, Bar>().apply {
  verify(myClass, times(2)).setItems(capture(0), capture(1))
  assertEquals("firstArgValue", firstValue[0].propOnFirstArg)
  assertEquals("secondArgValue", firstValue[1].propOnSecondArg)
  ...
}
```
Thus easily create multiple KArgumentCaptor instances for different types, capturing different arguments.

Currently I have code like this that I'd like to improve:
```
val nameUpdatedEventCaptor = argumentCaptor<NameUpdatedV1>()
        val streamNameCaptor = argumentCaptor<String>()
        verify(eventBusMock, times(1)).post(nameUpdatedEventCaptor.capture(), streamNameCaptor.capture())

        assertNotNull(nameUpdatedEventCaptor.firstValue)
        assertEquals(command.uuid, nameUpdatedEventCaptor.firstValue.personId)
        assertEquals(command.name, nameUpdatedEventCaptor.firstValue.name)
        assertEquals(
            NameService.TOPIC_NAME,
            streamNameCaptor.firstValue,
            "Event not sent to the expected topic"
        )
```

As you can see above, I have to declare multiple KArgumentCaptor, which is kind pretty verbose.

checketts:
This has been bothering me too. Here is my current solution:

```
  inline fun <reified A: Any,reified B: Any> captor2(): Pair<KArgumentCaptor<A>, KArgumentCaptor<B>> {
    return Pair(KArgumentCaptor(ArgumentCaptor.forClass(A::class.java), A::class), KArgumentCaptor(ArgumentCaptor.forClass(B::class.java), B::class))
  }

  inline fun <reified A: Any,reified B: Any, reified C: Any> captor3(): Triple<KArgumentCaptor<A>, KArgumentCaptor<B>, KArgumentCaptor<C>> {
    return Triple(
      KArgumentCaptor(ArgumentCaptor.forClass(A::class.java), A::class),
      KArgumentCaptor(ArgumentCaptor.forClass(B::class.java), B::class),
      KArgumentCaptor(ArgumentCaptor.forClass(C::class.java), C::class))
  }
```

So using this with @dsebastien example would look like:

```
captor2<Foo, Bar>().apply { foo, bar ->
  verify(myClass, times(2)).setItems(foo.capture(), bar.capture())
[...]

PR #300, opened 2018-10-13
Add argumentCaptor that create many at once for easy destructuring addresses #233

checketts:
This addresses #233 and provides an helper to create up to 5 `argumentCaptors` at once


nhaarman:
What about changing the signature to 

```
inline fun <reified A : Any, reified B : Any> argumentCaptor(
    a: KClass<A> = A::class,
    b: KClass<B> = B::class
)
```

and

```
inline fun <reified A : Any, reified B : Any, reified C : Any> argumentCaptor(
    a: KClass<A> = A::class,
    b: KClass<B> = B::class,
    c: KClass<C> = C::class
)
```

And so on? This avoids the `2`, `3` in the names.

Issue #328, opened 2019-02-26
Warning with `refEq`


pablisco:
We are testing something that uses `OkHttp` interceptors and came across an issue that we are not sure how to fix. We have this:

``` kotlin
verify(chain).proceed(refEq(expectedRequest))
```

To make sure the `Chain` is called with the right request. However, this gives us a `Type Mismatch` warning because It expects a `Request` and not `Request?` as `refEq` returns. 

The only thing that we have tried to do is to add `!!` at the end to remove the warning but that has woese consequences as it makes the tests fail with a `KotlinNullPointerException`. Which is understandable since `Mockito` returns `null` in their implementation.

Would it be as simple as using `?: createInstance()` as it is done on other matchers?

nhaarman:
Fixed in #329

PR #329, opened 2019-02-27
Add reified support for refEq like other matchers

pablisco:
This takes care of #328 

Follow the same pattern as with other matchers using `reified` generics to generate an instance to avoid having a warning or failure when using `!!`


pablisco:
@nhaarman
Any thoughts on this? It's been 4 months...

nhaarman:
Oops! I'll have a look at it soon.

nhaarman:
Thanks! Expect this in 2.2.0.

PR #357, opened 2019-08-31
Add suspend answer; Fix #346

### LINKED ISSUE #346: Support suspend blocks in Answers
When testing coroutines, I'd like to be able to mock a dependency with a `suspend` function with a custom `Answer` such that it delays, suspends indefinitely, or has other custom behavior besides simply returning a value. The particular test case I have in mind is testing that when the `CoroutineScope` of the class under test gets cancelled while one of its dependencies is suspending, it properly cancels the dependency and avoids further execution.

Example:
```Kotlin
runBlocking {
    whenever(repository.loadStatus()).willAnswer {
        // Suspend indefinitely
        suspendCancellableCoroutine<Status> {} <-- Compiler error 'Suspension functions can be called only within coroutine body'
    }
}
```
or
```Kotlin
mock<Repository> {
    onBlocking {
        loadStatus()
    } doAnswer {
        // Suspend indefinitely
        suspendCancellableCoroutine {} <-- Compiler error 'Suspension functions can be called only within coroutine body'
    }
}
```

MockK currently supports something like this with:
```Kotlin
coEvery {
    repository.loadStatus()
} coAnswers {
    // Suspend indefinitely
    suspendCancellableCoroutine<Status> {}
}
```

Testing asynchronous coroutines code is now achievable with the latest version of https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/.

neworld:
This code was inspired by:
- Java wrapper part: https://github.com/square/retrofit/blob/8c93b59dbc57841959f5237cb141ce0b3c18b778/retrofit/src/main/java/retrofit2/HttpServiceMethod.java#L168
- Kotlin wrapper part: https://github.com/square/retrofit/blob/8c93b59dbc57841959f5237cb141ce0b3c18b778/retrofit/src/main/java/retrofit2/KotlinExtensions.kt#L83-L98

I used Java code because kotlin does not allow call suspend functions or lambdas by giving own Continuation. But Kotlin does it under the hood, so Java could do it as well. For example:

```kotlin
fun foo(continuation: Continuation) {
  bar(continuation) //compilation error
}

suspend fun bar() {

}  
```

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/OngoingStubbing.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/OngoingStubbing.kt
@@ -26,10 +26,12 @@
 package org.mockito.kotlin
 
 import org.mockito.Mockito
+import org.mockito.internal.invocation.InterceptedInvocation
 import org.mockito.invocation.InvocationOnMock
 import org.mockito.stubbing.Answer
 import org.mockito.stubbing.OngoingStubbing
-import kotlin.DeprecationLevel.ERROR
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.intrinsics.startCoroutineUninterceptedOrReturn
 import kotlin.reflect.KClass
 
 
@@ -123,4 +125,16 @@ infix fun <T> OngoingStubbing<T>.doAnswer(answer: Answer<*>): OngoingStubbing<T>
  */
 infix fun <T> OngoingStubbing<T>.doAnswer(answer: (InvocationOnMock) -> T?): OngoingStubbing<T> {
     return thenAnswer(answer)
+}
+
+infix fun <T> OngoingStubbing<T>.doSuspendableAnswer(answer: suspend (InvocationOnMock) -> T?): OngoingStubbing<T> {
+    return thenAnswer {
+        //all suspend functions/lambdas has Continuation as the last argument.
+        //InvocationOnMock does not see last argument
+        val rawInvocation = it as InterceptedInvocation
+        val continuation = rawInvocation.rawArguments.last() as Continuation<T?>
+
+        // https://youtrack.jetbrains.com/issue/KT-33766#focus=Comments-27-3707299.0-0
+        answer.startCoroutineUninterceptedOrReturn(it, continuation)
+    }
 }
--- a/mockito-kotlin/src/test/kotlin/test/CoroutinesTest.kt
+++ b/mockito-kotlin/src/test/kotlin/test/CoroutinesTest.kt
@@ -7,8 +7,12 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.actor
+import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.mockito.kotlin.*
+import java.util.*
 
 
 class CoroutinesTest {
@@ -157,11 +161,106 @@ class CoroutinesTest {
             verify(testSubject).suspending()
         }
     }
+
+    @Test
+    fun answerWithSuspendFunction() = runBlocking {
+        val fixture: SomeInterface = mock()
+
+        whenever(fixture.suspendingWithArg(any())).doSuspendableAnswer {
+            withContext(Dispatchers.Default) { it.getArgument<Int>(0) }
+        }
+
+        assertEquals(5, fixture.suspendingWithArg(5))
+    }
+
+    @Test
+    fun inplaceAnswerWithSuspendFunction() = runBlocking {
+        val fixture: SomeInterface = mock {
+            onBlocking { suspendingWithArg(any()) } doSuspendableAnswer {
+                withContext(Dispatchers.Default) { it.getArgument<Int>(0) }
+            }
+        }
+
+        assertEquals(5, fixture.suspendingWithArg(5))
+    }
+
+    @Test
+    fun callFromSuspendFunction() = runBlocking {
+        val fixture: SomeInterface = mock()
+
+        whenever(fixture.suspendingWithArg(any())).doSuspendableAnswer {
+            withContext(Dispatchers.Default) { it.getArgument<Int>(0) }
+        }
+
+        val result = async {
+            val answer = fixture.suspendingWithArg(5)
+
+            Result.success(answer)
+        }
+
+        assertEquals(5, result.await().getOrThrow())
+    }
+
+    @Test
+    fun callFromActor() = runBlocking {
+        val fixture: SomeInterface = mock()
+
+        whenever(fixture.suspendingWithArg(any())).doSuspendableAnswer {
+            withContext(Dispatchers.Default) { it.getArgument<Int>(0) }
+        }
+
+        val actor = actor<Optional<Int>> {
+            for (element in channel) {
+                fixture.suspendingWithArg(element.get())
+            }
+        }
+
+        actor.send(Optional.of(10))
+        actor.close()
+
+        verify(fixture).suspendingWithArg(10)
+
+        Unit
+    }
+
+    @Test
+    fun answerWithSuspendFunctionWithoutArgs() = runBlocking {
+        val fixture: SomeInterface = mock()
+
+        whenever(fixture.suspending()).doSuspendableAnswer {
+            withContext(Dispatchers.Default) { 42 }
+        }
+
+        assertEquals(42, fixture.suspending())
+    }
+
+    @Test
+    fun willAnswerWithControlledSuspend() = runBlocking {
+        val fixture: SomeInterface = mock()
+
+        val job = Job()
+
+        whenever(fixture.suspending()).doSuspendableAnswer {
+            job.join()
+            5
+        }
+
+        val asyncTask = async {
+            fixture.suspending()
+        }
+
+        job.complete()
+
+        withTimeout(100) {
+            assertEquals(5, asyncTask.await())
+        }
+    }
 }
 
 interface SomeInterface {
 
     suspend fun suspending(): Int
+    suspend fun suspendingWithArg(arg: Int): Int
     fun nonsuspending(): Int
 }
 
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -31,7 +31,7 @@ dependencies {
     testCompile 'com.nhaarman:expect.kt:1.0.0'
 
     testCompile  "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    testCompile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.0.0'
+    testCompile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.0'
 
     testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.0.0"
 }


nhaarman:
Thanks! Looks like a useful feature :raised_hands: 

I'm not _that_ familiar with coroutines yet, does this solution rely on any undocumented Kotlin internals? I.e. is there any chance this may break in upcoming Kotlin versions, and if so why?

neworld:
Yes, it is. I am using it for some tests; otherwise, I was not able to write at all.

> does this solution rely on any undocumented Kotlin internals? 

I would say yes. Implementation details are not documented well. 

> is there any chance this may break in upcoming Kotlin versions

I believe this feature is quite safe:

1. The way suspend functions are working is compiler feature to add `Continuation` argument to all suspend functions. Kotlin devs should ensure backward compatibility. Otherwise, at some point, you could not upgrade kotlinc until all dependencies are rebuilt with a new compiler. 
2. Retrofit is using the same approach. They don't need weird java wrapper only because their code is in java already.

I suppose there is a little chance [the riskiest line](https://github.com/nhaarman/mockito-kotlin/pull/357/files#diff-ffbe41394337568b0bd6c05277fb4fb4R138) could not compile with future version.

neworld:
Created an issue: https://youtrack.jetbrains.com/issue/KT-33766

neworld:
@nhaarman, friendly ping. Any news for this solution? At Vinted we are using this solution for multiple cases and they are working well.

neworld:
Thanks to Roman Elizarov. He provided me with missing details and I was able to find a solution without Java.

mockito-kotlin/src/test/kotlin/test/CoroutinesTest.kt
```diff
@@ -157,11 +157,107 @@ class CoroutinesTest {
...
+            job.join()
+            5
+        }
+
+        val asyncTask = async {
+            fixture.suspending()
+        }
+
+        //TODO: on coroutines >1.3 it should be job.complete()
+        job.cancel()
```
bohsen: Maybe bump version of coroutines to latest version while at it and make this change. I'm confident the tests will pass as I've done the update on another branch earlier.

@nhaarman What do you think?

bohsen: Otherwise really nice job @neworld. Super simple implementation and the use case for this has been increasingly in demand since Android started supporting coroutine and since coroutines-test appeared.

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
```diff
@@ -124,3 +126,14 @@ infix fun <T> OngoingStubbing<T>.doAnswer(answer: Answer<*>): OngoingStubbing<T>
 infix fun <T> OngoingStubbing<T>.doAnswer(answer: (InvocationOnMock) -> T?): OngoingStubbing<T> {
     return thenAnswer(answer)
 }
+
+infix fun <T> OngoingStubbing<T>.willAnswer(answer: suspend (InvocationOnMock) -> T?): OngoingStubbing<T> {
```
mhernand40: Thoughts on naming this `doAnswer`? The compiler will consider it as a valid overload I believe.

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
```diff
@@ -26,10 +26,12 @@
 package com.nhaarman.mockitokotlin2
 
 import org.mockito.Mockito
+import org.mockito.internal.invocation.InterceptedInvocation
 import org.mockito.invocation.InvocationOnMock
 import org.mockito.stubbing.Answer
 import org.mockito.stubbing.OngoingStubbing
-import kotlin.DeprecationLevel.ERROR
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.intrinsics.startCoroutineUninterceptedOrReturn
```
mhernand40: I believe this has now been moved to an `*.experimental.*` package in the latest version of Kotlin.

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental.intrinsics/start-coroutine-unintercepted-or-return.html

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
mhernand40: I believe the `willAnswer` variants are extensions for `BDDMyOngoingStubbing`. May I request that we also add `suspend` support for `BDDMyOngoingStubbing`? 🙂

The tests would look something like:
```Kotlin
...
given(fixture.suspendingWithArg(any())).willAnswer {
    withContext(Dispatchers.Default) { it.getArgument<Int>(0) }
}
...
```
Thank you!

JustinBis:
Ping @nhaarman, I'd love to use this in my projects as well as an official feature. Thanks @neworld for the awesome PR.

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
connected-rmcleod: It looks like it has since been moved back again: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.intrinsics/start-coroutine-unintercepted-or-return.html

connected-rmcleod:
@nhaarman Adding to the friendly pinging. Code like this is fundamentally difficult to test without a suspending answer:
```
class ViewModel(private val api: API) {
  var isLoading: Boolean = false

  suspend loadData(): Data {
    isLoading = true
    val data = api.fetchData() // suspending call
    isLoading = false
    return data
  }
}
```
If I can't stub `fetchData` in such a way that it doesn't immediately return, it's pretty much impossible for me to write a passing test asserting that `isLoading` is ever true. In fact, the only way I could figure out how to do it was by copying and pasting the code from this PR into my code (or to use something other than Mockito.) 

I will add that while this PR uses low-level code, it is **not** marked experimental (it was briefly in an experimental package at one point but quickly moved back) and it **is** in the documentation: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.intrinsics/start-coroutine-unintercepted-or-return.html

It was also the approach suggested by the team lead for Kotlin libraries in the YT issue: https://youtrack.jetbrains.com/issue/KT-33766

Would be happy to help if there's any way I can assist in getting @neworld's excellent work merged here.

mhernand40:
FWIW, in case anyone feels blocked because of this limitation, I have actually begun testing without Mockito whenever possible by leveraging interfaces and this "fake" constructor pattern I have seen being used in many well known Kotlin libraries including the Kotlinx Coroutines library. Using the `fetchData()` example mentioned above by @connected-rmcleod:
```Kotlin
interface Api {
    suspend fetchData(): Data

    companion object {
        operator fun invoke(... /* dependencies */): Api {
            return ApiImpl(...)
        }
    }
}

private class ApiImpl(... /* dependencies */) {
    override suspend fetchData(): Data {
        // Implementation details here
    }
}
```
Or if you don't like the `companion object` allocation:
```Kotlin
@Suppress("FunctionName")
fun Api(... /* dependencies */): Api {
    return ApiImpl(...)
}

interface Api {
    suspend fetchData(): Data
}

private class ApiImpl(... /* dependencies */) {
    override suspend fetchData(): Data {
        // Implementation details here
    }
}
```

In code, I can still "instantiate" `Api` using `val api = Api(...)` and in tests I can create an inner `TestApi` and back it with a `CompletableDeferred` like so:
```Kotlin
    private val deferredData = CompletableDeferred<Data>()
    private val api = TestApi()

    ...

    private inner class TestApi() : Api {
        override suspend fetchData(): Data {
            return deferredData.await()
        }
    }
```
If I want to test the scenario where the fetch suspends indefinitely, I simply avoid calling `deferredData.complete(...)`. If I want the fetch to return a result immediately, I simply call `deferredData.complete(expectedData)`.

I get that this doesn't actually address the need for supporting suspend Answers in Mockito Kotlin but I figured I'd share how I have adapted my testing practices to some of the limitations. 🙂 

Examples of this "fake" constructor pattern:
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope.html
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job.html
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-supervisor-job.html
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-list.html

Notice how these are actually `fun`s.

neworld:
Actually you do not need to wait for the release. Thanks to kotlin extensions you could just copy-paste into your project:

```kotlin
/**
 * This should be replaced after the release of https://github.com/nhaarman/mockito-kotlin/pull/357
 */
@Suppress("UNCHECKED_CAST")
infix fun <T> OngoingStubbing<T>.willAnswer(answer: suspend (InvocationOnMock) -> T?): OngoingStubbing<T> {
    return thenAnswer {
        //all suspend functions/lambdas has Continuation as the last argument.
        //InvocationOnMock does not see last argument
        val rawInvocation = it as InterceptedInvocation
        val continuation = rawInvocation.rawArguments.last() as Continuation<T?>

        answer.startCoroutineUninterceptedOrReturn(it, continuation)
    }
}
```

I am using the same snippet for a year and works perfectly.

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
neworld: Yes. The same code was working for all coroutine and kotlin versions. Now I am using `1.4.0` with `1.3.9` of coroutines.

connected-rmcleod:
@mhernand40 @neworld All good advice! I'm aware that I can just roll my own test doubles or copy in the extension function. (In fact, in a lot of ways I prefer not using a mocking framework at all, but I've never found a team that agrees with me. xD) I'm just trying to emphasize that if someone is trying to write tests using mockito-kotlin and runs into this issue that it's a substantial stumbling block for them without having to resort to a different approach entirely.

neworld:
To be honest, recently I am trying to avoid mocking my own code at all. The topic was discussed many times and [there is plenty of good arguments](https://testing.googleblog.com/2013/05/testing-on-toilet-dont-overuse-mocks.html). The exception is for 3rd party libraries I have no control.

connected-rmcleod:
@neworld Without getting too much into the weeds, I think the correct position is between those two extremes, but I agree the far more common mistake is over-mocking. Over-mocking is a cancer.

connected-rmcleod:
This PR has been open for a long time. The approach also seems quite stable, and would be a pretty useful addition. Is there a substantial reason not to merge it in?

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
nickallendev: You might consider replacing startCoroutineUninterceptedOrReturn with it's implementation if the definition location isn't very stable. All it does is cast to expose the continuation parameter and invoke.
```
(answer as ((InvocationOnMock, Continuation<T?>) -> Any)).invoke(it, continuation)
```

TimvdLippe:
The mockito-kotlin artifact recently moved to the GitHub Mockito organization and we have been busy at work with publishing to Maven Central. Now that the dust mostly has settled, we can take a look at community PRs 😄   I will put this on my review list for this week, but I will have to get accustomed to the code a bit to understand what is going on here. From a first glance this looks fine, but I just want to make sure I understand what the overall logic is.

neworld:
Finished rebasing. 

> I just want to make sure I understand what the overall logic is

Suspend function under the hood returns `Any` object which could be [COROUTINE_SUSPENDED](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.intrinsics/-c-o-r-o-u-t-i-n-e_-s-u-s-p-e-n-d-e-d.html) in case of suspension. AFAIK all coroutines are started via `startCoroutineUninterceptedOrReturn` or similar primitives, which properly handles the return of the suspendable body. This block will be executed by another coroutine so it is important to properly execute the body:

https://github.com/mockito/mockito-kotlin/pull/357/files#diff-d34d30ea6bdd552f84d1ab3ea1e7fb5691080744548cdb5c0b84ca3c0fb78237R132-R137

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/OngoingStubbing.kt
```diff
@@ -124,3 +126,15 @@ infix fun <T> OngoingStubbing<T>.doAnswer(answer: Answer<*>): OngoingStubbing<T>
 infix fun <T> OngoingStubbing<T>.doAnswer(answer: (InvocationOnMock) -> T?): OngoingStubbing<T> {
     return thenAnswer(answer)
 }
+
+infix fun <T> OngoingStubbing<T>.willAnswer(answer: suspend (InvocationOnMock) -> T?): OngoingStubbing<T> {
```
TimvdLippe, line 130: The `willX` notation is already used in BDD-style mocking. To keep the naming consistent with the other methods defined, WDYT about the following: `doAnswerForCoroutine`?

mockito-kotlin/src/test/kotlin/test/CoroutinesTest.kt
TimvdLippe: Yeah I think this would be a good choice. Let's update to the latest supported syntax, which in this case is `job.complete()` as I understand it? (Not an expert here, so let me know if it has changed in the mean time)

TimvdLippe (approved):
Sorry for the long wait. I took a look at the PR and additional context and this solution makes sense to me. There were only 2 nits that would be great to be addressed and after that we can merge this.

Thanks @neworld for driving this forward and all others who have chimed in. Once this PR lands and I got through the rest of the PRs, I will publish a new version to Maven Central.

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/OngoingStubbing.kt
neworld: `doAnswerForCoroutine` sounds weird to me, but consistency is a first-class citizen always. Just maybe we should call it `doSuspendableAnswer`? It will sound like `on loadData do suspendable answer`

TimvdLippe: `doSuspendableAnswer` sounds good to me.

neworld: Final touched done. I leave `doSuspendableAnswer` name

mockito-kotlin/src/main/kotlin/com/nhaarman/mockitokotlin2/OngoingStubbing.kt
neworld: @TimvdLippe what do you think about this request? I could create `willSuspendableAnswer` for BDD as well.

TimvdLippe: We currently do not have any extensions for `BDDMyOngoingStubbing` in the existing `OngoingStubbing.kt`. I think we should treat that as a separate improvement, but I would happy to approve a PR on that. For now, let's keep that separate and merge this PR as-is.

TimvdLippe (approved):
Thank you so much for your continued effort and apologies that it took so long! I will make sure this change will be published to Maven Central after a successful build.

TimvdLippe:
This feature is available per version 3.1.0: https://repo1.maven.org/maven2/org/mockito/kotlin/mockito-kotlin/3.1.0/

PR #360, opened 2019-09-03
Fix Travis

BraisGabin:
CI is failing because the new images of travis doesn't work with `oraclejdk8`. For this reason I moved to `openjdk8`


nhaarman:
Thanks!

Issue #372, opened 2020-01-15
Mocking an interface with non-nullable properties sets the properties as null


JordiGarcL:
It looks like Mockito is using `null` as a default value for properties in a interface even though they are non-nullable.

Reproduce: 
```kotlin
interface Dog {
	val name: String
}

class DogCaller {
	fun callDog(dog: Dog) = "Hey ${dog.name} come here!"
}


class DogCallerTests {

	private val dogCaller = DogCaller()

	@Test
	fun `dog caller should call dog by its name`() {
		val beagle: Dog = mock()

                // This makes the test pass                
                // whenever(beagle.name).thenReturn("Firulais")

		dogCaller.callDog(beagle)

		verify(beagle).name // Fails because it detects that dog.name : String = null
	}
}
```

bohsen:
Your reproducer passes when I try it out. Maybe try checking your imports.

Also returning null is the expected behaviour when a method hasn't been stubbed.

JordiGarcL:
Sorry, now I see I got this error cause inside `callDog()` I passed `dog.name` to another function which expected a non-nullable type. That's where the nullability of `String` check failed. 

Though from  the`castNull()` in the file you referenced I see this is a totally intentional behaviour. Should it be the case though? `String` has always been a "special" type. I expected the mock to return a '"nice" value such as `""`.

bohsen:
It would make sense to handle it that way. Would actually solve the issue seen in #241 without using the proposed workaround I believe. `String`s have always been kind of problematic in regards to Mockito-kotlin.

PR #373, opened 2020-01-15
Return an empty String instead of null when creating a mock

jordigarcl:
See #372 for reference

I believe `String` values in mocks should return `""` by default.

It is a historically 'special' type which has always had some special 'primitive-like' privileges. 
I assume people expect it to return a nice value as well!


JordiGarcL:
Has this PR been reviewed yet?

wesalvaro:
What does plain Mockito do?

wesalvaro:
Plain Mockito returns [`null`](https://github.com/mockito/mockito/blob/main/src/main/java/org/mockito/internal/util/Primitives.java).

jselbo:
Closing because Mockito-Kotlin relies on Mockito for this behavior. You might be interested in using `Mockito.RETURNS_SMART_NULLS`

https://www.javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#RETURNS_SMART_NULLS

PR #377, opened 2020-03-13
Add additional matchers

tschuchortdev:
Adds some additional matchers like `not`, or `leq`. I couldn't be bothered to write tests because the code is so incredibly simple.


perkss:
Any plans to put this in?

TWiStErRob:
@tschuchortdev the code is simple, but it's still code, and the tests would protect it from breaking in the future. For example, if the functions get reordered, and accidentally change bodies or something we don't know yet. Verifying a basic requirement that the function delegates to the right one (by validating what it does e.g. tests would be different for lt and leq).

lgladysz:
I think this is already merged in #508

TWiStErRob:
agreed

PR #388, opened 2020-09-23
Fixed argThat cannot write matcher for base types

deanssss:

 - [v] Can you back your code up with tests?
 - [v] Keep your commits clean: [squash your commits if necessary](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).


wesalvaro:
Would be better if there was a test or more explanation about what this fixes.

deanssss:
> Would be better if there was a test or more explanation about what this fixes.

I opened this PR many years ago and have forgotten the details of it now. 
I'll close it now, maybe reopen it in the future if I remember it.

PR #405, opened 2021-01-27
Enabled releases to GitHub and Bintray

mockitoguy:


TimvdLippe (approved):
After this we can do a practice release to Maven Central right?

PR #414, opened 2021-03-22
fix: package groupId in install section

mariorez:
This update corrects the `groupId` path, needed to install the library successfully.


TimvdLippe:
@mockitoguy could you confirm that this new location is intended or that we should update our artifact location in Maven Central?

TimvdLippe (approved):
Confirmed privately this is the correct location. Thanks for updating teh readme!

PR #415, opened 2021-03-23
ignore all intelliJ files

hennr:
Ignore intelliJ files in an easier way.
I stumbled upon this as IntelliJ watned to commit jarRepositories.xml on my machine, this commit ignores this file as well.


TimvdLippe (approved):
Yup, thanks!

Issue #400, opened 2021-01-05
Migrate release process to Shipkit


TimvdLippe:
Similar to `mockito-core` and `mockito-scala`, let's migrate the release process to Shipkit. The Maven Central artifact location should become https://repo1.maven.org/maven2/org/mockito/mockito-kotlin/

mockitoguy:
Sure!

TimvdLippe:
@mockitoguy I would like to finish this task at the end of this week, so that we can prepare the announcements for the artifact move and do some general cleanups. Would it be possible if you could get us a release on Maven Central by this week or is that too ambitious?

mockitoguy:
I'll put some cycles into it this week. Let's see how it goes. Thanks!

mockitoguy:
I'm making progress in this branch: https://github.com/mockito/mockito-kotlin/tree/sf Thanks for patience!

TimvdLippe:
Fixed 🎉

hennr:
It's not on mvn central yet though, is it?

TimvdLippe:
It's published at https://repo1.maven.org/maven2/org/mockito/kotlin/mockito-kotlin/

hennr:
Yep, does as well for me now. Nice!

Let's see if search.maven.org finds it as well tomorrow.

hennr:
I just included the new artifact in our project, works fine!

PR #416, opened 2021-03-23
correct mvn badge

hennr:
This resolves my confusion about not finding the artifact on central expressed in #400


TimvdLippe (approved):
Ah yes, good catch!

PR #420, opened 2021-03-23
migrate to JUnit5

hennr:
Not sure if you want to, but why not migrate to JUnit 5 I thought :)

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -95,7 +95,6 @@
       <option name="XML_KEEP_LINE_BREAKS" value="false" />
       <option name="XML_ALIGN_ATTRIBUTES" value="false" />
       <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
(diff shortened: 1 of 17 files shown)


TimvdLippe:
Ideally I would like to do this when we switch over the main Mockito test suite (https://github.com/mockito/mockito/issues/1246) and I don't think we want to move to JUnit 5 just yet. Let's discuss further in the Mockito issue before we merge any PRs related to JUnit 5.

hennr:
All right, no worries. If you need a helping hand on the other test code, let me know.

bohsen:
@hennr Just curious. What are the advanteges of using junit5 over junit4 (with respect to an updated junit dependency like junit 4.13.2)?

TimvdLippe:
@bohsen That would be the same question for me, but let's move that discussion to the Mockito issue rather than this PR, so that we can keep this PR focused on the implementation. Thanks!

niki-belo-scacap:
You folks might consider closing this one - it's been a while

PR #423, opened 2021-03-29
Add required tags argument to release push command

TimvdLippe:
Without `--tags`, the push command will state:
```
Everything up-to-date
```

Instead, we have to explicitly list the `--tags` argument to
push tags to the remote.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -5,7 +5,7 @@ to Sonatype snapshot repo at https://s01.oss.sonatype.org/content/repositories/s
 2. In order to release a non-snapshot version to Maven Central push an annotated tag, for example:
 ```
 git tag -a -m "Release 3.4.5" 3.4.5
-git push
+git push --tags
 ```
 3. At the moment, you **may not create releases from GitHub Web UI**.
 Doing so will make the CI build fail because the CI creates the changelog and posts to GitHub releases.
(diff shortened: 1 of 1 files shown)


mockitoguy:
Absolutely!

RELEASING.md
```diff
@@ -5,8 +5,8 @@ to Sonatype snapshot repo at https://s01.oss.sonatype.org/content/repositories/s
 2. In order to release a non-snapshot version to Maven Central push an annotated tag, for example:
 ```
 git tag -a -m "Release 3.4.5" 3.4.5
-git push
+git push --tags
```
mockitoguy, line 8: How about `git push origin 3.4.5`? It's a little safer because it pushes an explicit tag. 

Great catch!

mockitoguy (approved):
Thanks!

RELEASING.md
TimvdLippe: Confirmed that the tag is pushed and the release is run: https://github.com/mockito/mockito-kotlin/runs/2219360718?check_suite_focus=true I am not sure if the binary has actually been pushed to Maven Central (the logs don't list anything), but that's a separate issue.

TimvdLippe: Also, the release notes look rather empty: https://github.com/mockito/mockito-kotlin/releases/tag/3.0.0 I will manually update these notes to include the relevant PRs.

mockitoguy: Thanks! I'll look into the release notes issue.

TimvdLippe: Seems like the artifacts were published as expected: https://repo1.maven.org/maven2/org/mockito/kotlin/mockito-kotlin/3.0.0/

PR #425, opened 2021-03-29
Run release procedure on tag pushes

TimvdLippe:
Running CI actions on tags requires tags to be explicitly listed per the
documentation in
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags

Also update the logic to make sure we run on tags, not just on the main
branch pushes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 3.*
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,6 +11,8 @@ on:
   push:
     branches:
       - main
+    tags:
+      - 3.*
```
mockitoguy, line 15: Good catch! I'm surprised it's needed.

mockitoguy (approved):
Nice!

PR #428, opened 2021-03-30
Update link to Sonatype snapshot repo

shestee:
Current link in RELEASING.md doesn't lead directly to 'mockito-kotlin' snapshots repo. This commit fixes the link with valid URL.


TimvdLippe (approved):
Good catch!

PR #429, opened 2021-04-01
Add support of suspendable answers to BDD

neworld:
Post suspendable answer support to BDD mockito as well.

#357


TimvdLippe (changes requested):
Can we mirror all other methods listed in `OngoingStubbing.kt` to mirror them to `BDDMockito.kt`?

neworld:
> Can we mirror all other methods listed in `OngoingStubbing.kt` to mirror them to `BDDMockito.kt`?

Yes sure. But I found only a few cases missing. All others are covered by mockito itself. Also, changed the commit name.

TimvdLippe (approved):
Thanks!

PR #440, opened 2021-09-05
Issue 439: added the KMemberAccessor which can be activated to better…

juan-palacios:
… handle Kotlin synthetic constructors and functions


juan-palacios:
Need to add a test case for the `invoke` path. I also need to fix the assumption of the two last arguments being markers. In method invocations, it behaves slightly differently with the 1st parameter being `this`

PR #447, opened 2021-10-08
Remove deprecated verifyZeroInteractions

### LINKED ISSUE #446: Remove verifyZeroInteractions?
I was trying to update to Mockito 4, and I noticed that I was getting `NoSuchMethodError`s when using Mockito-Kotlin's [`verifyZeroInteractions`](https://github.com/mockito/mockito-kotlin/blob/3.2.0/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Verification.kt#L84-L91).

After looking into it, it turns out that this was happening because `Mockito.verifyZeroInteractions` was [removed as of Mockito 4](https://github.com/mockito/mockito/pull/2418), and Mockito-Kotlin hasn't been updated to build against Mockito 4 yet. I also noticed that `verifyZeroInteractions` has been [marked as deprecated](https://github.com/mockito/mockito-kotlin/pull/401) in Mockito-Kotlin, but there hasn't been a new release since then.

If there are plans of releasing a Mockito-Kotlin 4 soon, maybe `verifyZeroInteractions` can be removed? Or if there's going to be another update for 3.x first, maybe the implementation of `verifyZeroInteractions` can be changed to call through to `verifyNoMoreInteractions` instead, so that it's usable (but still deprecated) with Mockito 4?

rashadsookram:
Fixes #446.

I just removed the method for now, but if it's preferred to keep it around and call through to `verifyNoMoreInteractions` instead, let me know and I can update the PR.

---

   - there were no tests for `verifyZeroInteractions` as far as I can tell

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -25,7 +25,7 @@ dependencies {
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.0.0'
 
-    compile "org.mockito:mockito-core:3.12.4"
+    compile "org.mockito:mockito-core:4.0.0"
 
     testCompile 'junit:junit:4.12'
     testCompile 'com.nhaarman:expect.kt:1.0.0'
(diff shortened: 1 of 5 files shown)


ratepay-sstein:
When can we expect a release of mockito-kotlin, which is compatible with Mockito 4?

TimvdLippe:
@ratepay-sstein I had pushed the 4.0.0 tag yesterday, but it seems like it wasn't running the release procedure. #448 should fix that, thanks for letting us know!

PR #450, opened 2021-11-17
Remove jcenter

k163377:
Removed `JCenter` from `build.gradle` as it is scheduled to be closed in February 2022.
Upgraded to `1.0.1` as `com.nhaarman:expect.kt:1.0.0` can no longer be obtained due to this change.
https://mvnrepository.com/artifact/com.nhaarman/expect.kt

The description of `README` has been fixed at the same time.

## Addition
No changes were made between `com.nhaarman:expect.kt:1.0.0` and `1.0.1` except for the `publish` change.
https://github.com/nhaarman/expect.kt/compare/1.0.0...1.0.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ A small library that provides helper functions to work with [Mockito](https://gi
 
 ## Install
 
-Mockito-Kotlin is available on Maven Central and JCenter.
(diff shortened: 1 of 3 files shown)


TimvdLippe (approved):
Awesome, thanks for fixing!

Issue #325, opened 2019-02-20
Add support for per-mock strictness


thevoiceless:
See https://github.com/mockito/mockito/pull/1272

Mockito 2.20.1 added support for per-mock/stubbing strictness. It looks like the most straightforward option would be passing an additional argument to `withSettings()`.

thevoiceless:
For anyone wanting to use this in the meantime: You simply replace `whenever` with `` Mockito.lenient().`when` ``, or` Mockito.mock(Foo::class.java, withSettings().lenient())` to make the entire mock lenient.

thevoiceless:
Actually, it looks like this may have been fixed this 10 days ago with https://github.com/nhaarman/mockito-kotlin/pull/353

PR #454, opened 2022-03-04
Add LenientStubber

taeyeon-Kim:


TimvdLippe (commented):
This looks good to me. Just to confirm, this is intended to fix #325 right?

taeyeon-Kim:
@TimvdLippe Actually... I created a PR because I needed it, but it seems to be related to #325 as well.

TimvdLippe (approved):
Thanks for the fix!

PR #455, opened 2022-03-14
use default argument instead of having 2 functions?

ericytsang:

    - all existing tests pass; the interface of the API didn't change, but there is less code to maintain now, too
 - [X] Keep your commits clean: [squash your commits if necessary](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
    - there is only one commit in this PR

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
@@ -35,55 +35,6 @@ import kotlin.DeprecationLevel.ERROR
 import kotlin.reflect.KClass
 
 /**
- * Creates a mock for [T].
- *
- * @param extraInterfaces Specifies extra interfaces the mock should implement.
- * @param name Specifies mock name. Naming mocks can be helpful for debugging - the name is used in all verification errors.
- * @param spiedInstance Specifies the instance to spy on. Makes sense only for spies/partial mocks.
- * @param defaultAnswer Specifies default answers to interactions.
- * @param serializable Configures the mock to be serializable.
- * @param serializableMode Configures the mock to be serializable with a specific serializable mode.
- * @param verboseLogging Enables real-time logging of method invocations on this mock.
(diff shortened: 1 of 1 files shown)


mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
```diff
@@ -112,7 +63,7 @@ inline fun <reified T : Any> mock(
     @Incubating useConstructor: UseConstructor? = null,
     @Incubating outerInstance: Any? = null,
     @Incubating lenient: Boolean = false,
-    stubbing: KStubbing<T>.(T) -> Unit
+    stubbing: KStubbing<T>.(T) -> Unit = {}
```
TimvdLippe, line 66: We use this value on line 84 and now we unconditionally create a KStubbing even if you didn't specify it. I am okay if we refactor this method to check for an empty value and then creates the KStubbing if it isn't. Can you adapt the PR to do thst?

TimvdLippe (changes requested):

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
ericytsang: Yes! I can do that. I'll make `stubbing` nullable, and give it a null default value, and only create the KStubbing if it is non-null..? Thanks for considering my PR

TimvdLippe: Yes that works for me. Thanks for the PR and looking forward to the fix 😄

TimvdLippe: Unfortunately, I think you are correct in that this would be a breaking change. I am not happy about the duplication either, but I don't think it is worth breaking our users for, given it is only a small burden for us. Therefore, I think it is better if we leave it as-is, unfortunately. Thanks for spotting this though!

PR #464, opened 2022-10-24
Ease verify inOrder for suspending functions

seyfahni:
Why?
```kotlin
val stub = mock<SomeInterface>()

runBlocking {
  testWith(stub) // testing suspending code
}

stub.inOrder {
  verify().notSuspending()
  // this "worked" before my change, but didn't verify order
  verifyBlocking { suspending() }
  // instead one had to use runBlocking { verify().suspending() }
  // or val v = verify(); runBlocking { v.suspending() }
}
```

Changes:
- Added interface `KInOrder` to allow extending mockito's `InOrder` interface:
  - defined `verifyBlocking` methods with similar definition to those in `Verification`
  - implemented interface with a decorator that takes a mockito `InOrder` and does blocking verification the same way as in `Verification`
- Adjusted return types in `Verification` to provide `KInOrder`:
  - the implementation (`KInOrderDecorator`) is not `internal` so that the `inline fun` declaration continues working
- Added methods to `InOrderOnType` that allow passing a `VerificationMode`.

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/KInOrder.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/KInOrder.kt
@@ -1,0 +1,47 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2018 Niek Haarman
+ * Copyright (c) 2007 Mockito contributors
+ *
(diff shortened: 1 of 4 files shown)


TimvdLippe (approved):
Nice fix!

PR #476, opened 2023-01-19
Exchange explanations between verifyNoInteractions() and verifyNoMoreInteractions()

tatsuyafujisaki:
The explanation for `verifyNoInteractions()` and the explanation for `verifyNoMoreInteractions()` seem to be exchanged. This pull request is to undo the exchange.


TimvdLippe (approved):
Good catch!

Issue #453, opened 2022-02-10
What about improving  doReturn -> when implementation?


htafoyag:
I believe there's a missed opportunity on the `doReturn.when` implementation. Mockito kotlin is only changing reseved word `'when' `with `whenever`, but the implementation is not on par with the rest of the library.

When using on` { myMethod() } doReturn "something"`,  myMethod is inferred from the mock scope.  However the doReturn is not doing the same and we need to resend the mock to whenever method:

```kotlin
val myMock = mock<MyClass> {
    doReturn("something").whenever(myMock).myMethod()
}
```

A better format would be:

```kotlin
val myMock = mock<MyClass> {
    doReturn("something").whenever(myMethod())
}
```

or even a format similar to the ones already there:

```kotlin
val myMock = mock<MyClass> {
    doReturn "something" whenever { myMethod() }
}
```

The same for `doNothing` and other similar methods

Or by a fact, i would've preferred  that `on - doReturn` called `doReturn.when`  instead,  I see no purpose on invoking the real methods and the need to use a different method for spies.

thecodewarrior:
The problem with `on { myMethod() } doReturn "something"` using the `doReturn("something").when(mock).myMethod()` internally is that there's no way to tell when the mock is "done"

At no point in the chain do we know that we're "done" and can wrap it up.
```kotlin
val myMock = mock<MyClass> {
    on { myMethod() }  doReturn "first" doReturn "second" doThrow IllegalStateException()
}
```
Technically we could store the current stub in the KStubbing and close the stubbing the next time `on` is called, and again at the end of the block, but that seems a bit iffy to me.

I would propose a different syntax, which provides symmetry with the existing implementation. It would be extremely easy to implement as well.

```kotlin
val myMock = mock<MyClass> {
    doReturn("first").on { myMethod() }
}
```

I went with `doReturn(...).on { ... }` instead of `doReturn(...) on { ... }` because of the difference in how they're chained. With the `on { ... } doReturn` syntax, everything is chained as infix functions, however with this syntax you have to chain with dot syntax. 
```kotlin
val myMock = mock<MyClass> {
    // consistent dot notation
    doReturn("first").doReturn("second").doThrow(IllegalStateException()).on { myMethod() }
    // no dot notation is inconsistent and awkward after chaining
    doReturn("first").doReturn("second").doThrow(IllegalStateException()) on { myMethod() }
}
```

Implementing it would just require one extra method in `KStubbing`:
```kotlin
[...]

PR #481, opened 2023-03-17
Add `doReturn().on { method() }` helper to KStubbing

thecodewarrior:
This implements my proposed solution to #453, which provides a cleaner way to use the `doReturn().when(mock).methodCall()` mocking syntax. Mockito-kotlin provides a `.whenever(mock)` extension, since `when` is a reserved keyword in Kotlin, however inside a `KStubbing` you still need to specify the instance, which gets messy and verbose. 

I added one extension function to `KStubbing`: `Stubber.on(methodCall)`. 

```kotlin
// before
val myMock = mock<MyClass> {
    doReturn("first").whenever(it).myMethod()
}
// after
val myMock = mock<MyClass> {
    doReturn("first").on { myMethod() }
}
```

I went with `doReturn(...).on { ... }` instead of `doReturn(...) on { ... }` because of the difference in how the results are chained:
```kotlin
val myMock = mock<MyClass> {
    // `on {} doReturn X doThrow Y`   - uniformly uses spaces
    // `doReturn(X).doThrow(Y).on {}` - uniformly uses dots
    // `doReturn(X).doThrow(Y) on {}` - inconsitent dots/spaces

    on { myMethod() } doReturn "value" doThrow IllegalStateException()
    doReturn("first").doReturn("second").doThrow(IllegalStateException()).on { myMethod() }
    doReturn("first").doReturn("second").doThrow(IllegalStateException()) on { myMethod() }
}
```

This method doesn't provide type safety, but it's a simple and effective solution. We could create an entire custom `KStubber<T>` class that does type safety, however, while that will provide type *safety*, it won't provide type *inference*, since you do the `.on { myMethod() }` at the *end*.

We could provide type safety *and* type inference given a different syntax, but it's awkward.
```kotlin
val myMock = mock<MyClass> {
    on({ myMethod() }) {
        doReturn("first")
        doReturn("second")
        doThrow(IllegalStateException())
    }
}
```

There are probably other possible syntaxes, but this is the best one I came up with.

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/KStubbing.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/KStubbing.kt
@@ -30,6 +30,7 @@ import kotlinx.coroutines.runBlocking
 import org.mockito.Mockito
 import org.mockito.exceptions.misusing.NotAMockException
 import org.mockito.stubbing.OngoingStubbing
+import org.mockito.stubbing.Stubber
 import kotlin.reflect.KClass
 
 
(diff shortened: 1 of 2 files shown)


TimvdLippe (approved):
Thanks!

Issue #478, opened 2023-02-06
Mockito 5.1 support


elomonaco:
Mockito is now on version 5.1 and it appears the kotlin version is still on Mockito 4.1. Is there any plans to get in sync with the main Mockito library?

Mockito 5.1 has a few things I'd like to use but this is a blocker.

TimvdLippe:
I would be happy to merge a PR that bumps the version of `mockito-core`

lukas-krecan:
[This](https://github.com/mockito/mockito-kotlin/pull/482) may be a good start.

Issue #474, opened 2023-01-19
Mockito 5.0 breaks vararg argument capture in Kotlin


emartynov:
Hence [issue](https://github.com/mockito/mockito/issues/2878) was wrongly put on the mockito project.

Have code: 

```
class PreferencesRepository {
    suspend fun getPreferences(
        vararg preferences: String
    ): Map<String, Any> 
{}
```

Write tests:
```
        val preferencesRepository: PreferencesRepository = mock()
        val preferencesCapture = argumentCaptor<String>()
        whenever(preferencesRepository.getPreferences(anyVararg())) doReturn emptyMap()

        verify(preferencesRepository).getPreferences(preferencesCapture.capture())
```

Run test and observe:
```
at PreferencesRepository.getPreferences(PreferecesRepository.kt:11)
Actual invocations have different arguments:
preferencesRepository.getPreferences(
    [String, String]
);
```

emartynov:
Moved from https://github.com/mockito/mockito/issues/2878

TimvdLippe:
If anybody would like to stab at this, I am happy to review a PR with a regression test + fix.

kschults:
I see that the PR 2807 got merged, which looks like it's trying to fix this. Any estimate on when this will be available?

kschults:
Oh, I see, I misread the timeline. Thanks.

samabcde:
Come up with this workaround. Honestly speaking, I don't really understand why it works.
```
inline fun <reified T : Any, reified E : Any> anyVararg(clazz: KClass<E>): T {
    return ArgumentMatchers.any(T::class.java)
          ?: arrayOf(createInstance(E::class)) as T
}

// below methods are from org/mockito/kotlin/internal/CreateInstance.kt
fun <T : Any> createInstance(kClass: KClass<T>): T {
    return castNull()
}
@Suppress("UNCHECKED_CAST")
private fun <T> castNull(): T = null as T
```
```
    @Test
    fun testVararg() {
        val preferencesRepository: PreferencesRepository = mock()
        val preferencesCapture = argumentCaptor<String>()
        // not the * before anyVararg
        whenever(preferencesRepository.getPreferences(*anyVararg(String::class))) doReturn emptyMap()
        preferencesRepository.getPreferences("works!")
        verify(preferencesRepository).getPreferences(preferencesCapture.capture())
        assert(preferencesCapture.lastValue == "works!")
    }
```

emartynov:
[...]

PR #482, opened 2023-05-19
#478 Upgrade Mockito 

lukas-krecan:


lukas-krecan:
It's not compatible with older Mockito versions so it should be a major version release.

emartynov:
Is it for #474 ?

lukas-krecan:
> Is it for #474 ?

Yes, and https://github.com/mockito/mockito-kotlin/issues/478

lukas-krecan:
Sorry, I do not understand. Which tests do you mean?

TimvdLippe:
Mockito 5 made breaking changes to varargs matchers. I expected that to have consequences for Kotlin as well. However, I see little to no changes in the regression test suite of `mockito-kotlin` in this PR. Does that mean that these breaking changes were not relevant for Kotlin folks, or that we are missing test suite coverage?

lukas-krecan:
> Does that mean that these breaking changes were not relevant for Kotlin folks, or that we are missing test suite coverage?

I guess both. The only impacted method here is `anyVararg`, I had to add a tests as I did not trust the current coverage. Moreover, in Kotlin you can't just use an array in vararg, you have to use the `method(*array)` syntax. Maybe it helped.

But I am a first-time contributor to this library, so taky me with a huge grain of salt.

lukas-krecan:
Just for completenes, changing the signature to `fun <T : Any> anyVararg(clazz: KClass<T>): Array<T>` and using it like this `method(*anyVararg())` might be cleaner, but would complicate the migration path without much additional value.

TimvdLippe (approved):
Sounds good to me, thanks!

Issue #480, opened 2023-03-03
lenient().whenever() reports type mismatch TypeVariable(T) was expected


ebrowne72:
I updated to 4.1.0 and replaced a ``lenient().`when`()`` with `lenient().whenever()`, but it gives the following error:

    Type mismatch: inferred type is String? but TypeVariable(T) was expected

PR #485, opened 2023-07-10
Fix issue #480 by widening the allowed types for lenient().whenever()

hunterwerlla:

 - [ x ] Can you back your code up with tests?
 - [ x ] Keep your commits clean: [squash your commits if necessary](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

Issue #480 is caused by constraining the type to `Any`. To fix this, we can update it to have no constraint like the underlying Java code


TimvdLippe (approved):
Thanks!

Issue #488, opened 2023-07-18
Version 5.0.0 + Android fails with: Cannot inline bytecode built with JVM target 11


bvschaik:
Android projects still use JVM target 1.8 due to device constraints; the gradle file therefore contains:

```
android {
    ...
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
    kotlinOptions {
        jvmTarget = '1.8'
    }
}
```

Upgrading mockito-kotlin from 4.1.0 to 5.0.0 causes the following compile error on any mock/whenever/etc call:

> Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option

JVM target 11 is nowhere to be found in the Gradle files.

Steps to reproduce:

1. Clone minimal example project: https://github.com/bvschaik/mockito-kotlin-error
2. Try to run unit tests: `./gradlew testDebugUnitTest`
3. Note compile errors

If you downgrade mockito-kotlin to 4.1.0, the project compiles and runs successfully.

If mockito-kotlin no longer supports JVM 1.8, that should've been noted in the changelog.

argenkiwi:
Does this mean we are stuck with v4.x for Android development for the foreseeable future?

TimvdLippe:
The drop of compatibility is listed on mockito-core in its release notes combined with the rationale: https://github.com/mockito/mockito/releases/tag/v5.0.0

If you still need to support Java 8, you can keep on using Mockito 4. Nothing should break there.

naitaku:
I was unaware that Mockito 5 no longer offers support for Java 8. Does this imply that for Android projects, the jvmTarget should now be set to VERSION_11? It seems to be supported with Android Gradle Plugin 7.0.
https://developer.android.com/build/releases/past-releases/agp-7-0-0-release-notes#java-11

If that's the case, would it be advisable to include a note in the Readme.md specifically for Android developers regarding this change?

TimvdLippe:
Yes that is indeed now required. Feel free to send me a PR with a README update on the version similar to what we have in the README of `mockito-core`.

wiryadev:
i have set jvmtarget and compatibility in compileoptions to Java 11 but the problem still exists

naitaku:
@wiryadev, would it be possible for you to share an example project that demonstrates the issue you're encountering? 

[...]

PR #493, opened 2023-09-19
#488: Change kotlinOptions.jvmTarget to VERSION_1_8

naitaku:
This commit addresses issue #488, where Android projects were encountering a build failure with the error message: "Cannot inline bytecode built with JVM target 11." 
The issue was resolved by adjusting the 'jvmTarget' to 'VERSION_1_8,' ensuring compatibility with Android projects. 

 * Add mockito-kotlin dependency with testImplementation and androidTestImplementation to an Android project, and mock function works.
 - [X] Keep your commits clean: [squash your commits if necessary](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -50,7 +50,7 @@ dokka {
 
 tasks.withType(KotlinCompile).configureEach {
     kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_11
+        jvmTarget = JavaVersion.VERSION_1_8
     }
 }
 


TimvdLippe:
Mockito 5 no longer supports Java 8. Therefore, I don't think we should do that for mockito-kotlin either, should we?

naitaku:
My understanding is that mockito-core 5.x no longer supports Java 8, and 'jvmTarget' should be set to 'VERSION_11.'
I'll close this PR.

PR #496, opened 2023-11-29
Main dependencies updated

Nynuzoud:
- gradle wrapper to 8.1
 - kotlin to 1.9.20
 - java to version 17
 - mockito-core to version 5.7.0
 - few other related changes

The idea is to get recent java version and keep the project up to date, so there are less issues with modern plugins that work on Java 17 i.e. AGP 8.x


mockito-kotlin/build.gradle
```diff
@@ -25,33 +26,34 @@ dependencies {
...
+            remoteLineSuffix.set("#L")
+        }
     }
 }
 
 tasks.withType(KotlinCompile).configureEach {
-    kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_11
+    compilerOptions {
+        jvmTarget.set(JvmTarget.JVM_17)
```
TimvdLippe, line 55: Mockito still supports Java 11, so we can't bump this here. Please revert the changes to Java 17, the rest looks good to me.

Nynuzoud:
I updated by downgrading to Java 11 but got a failing test:
`doReturn_withGenericIntReturnType_on`

Will take a look a bit later. Updating `Strictness` doesn't help

tests/src/test/kotlin/test/OngoingStubbingTest.kt
```diff
@@ -288,14 +297,13 @@ class OngoingStubbingTest : TestBase() {
...
-            mock<GenericMethods<Int>> {
-                on { genericMethod() } doReturn 2
-            }
+        /* When */
+        val mock = mock<GenericMethods<Int>> {
+            on { genericMethod() } doReturn 2
         }
+
+        /* Then */
+        expect(mock.genericMethod()).toBe(2)
```
Nynuzoud, line 306: seems like it is not crashing anymore here with Generic methods

PR #498, opened 2023-12-02
Increase visibility of Mockito dependency

### LINKED ISSUE #497: Regression in 5.2.0 POM: missing Mockito dependency for compilation
## Symptom
```
e: ...kt:0:0 Cannot access class 'org.mockito.stubbing.Stubber'. Check your module classpath for missing or conflicting dependencies
e: ...kt:0:0 Cannot access class 'org.mockito.invocation.InvocationOnMock'. Check your module classpath for missing or conflicting dependencies
e: ...kt:0:0 Unresolved reference: callRealMethod
```

## Repro
```kotlin
interface Factory {
    fun create(): String
}

@Test fun test() {
	val mock = mock<Factory>()
	doAnswer { it.callRealMethod() } // More complex IRL, just simplified, because it doesn't matter.
		.whenever(mock)
		.create()
}
```
image
Note: somehow by simplifying Kotlin knows what `InvocationOnMock` & `callRealMethod` is, but it shouldn't, anyway `whenever` is still broken.

## Issue
Compare:
https://repo1.maven.org/maven2/org/mockito/kotlin/mockito-kotlin/5.1.0/mockito-kotlin-5.1.0.pom
https://repo1.maven.org/maven2/org/mockito/kotlin/mockito-kotlin/5.2.0/mockito-kotlin-5.2.0.pom

Diff
```diff
 <dependency>
     <groupId>org.mockito</groupId>
     <artifactId>mockito-core</artifactId>
-    <version>5.3.1</version>
+    <version>5.7.0</version>
-    <scope>compile</scope>
+    <scope>runtime</scope>
 </dependency>
```

## Background

Root cause is this change:
https://github.com/mockito/mockito-kotlin/pull/496/files#diff-09728e1a022bed42cb4f8d1d2418235341d939c7f641a015f30648911ec97cc4R29

When upgrading Gradle from `compile` to `implementation` @Nynuzoud didn't make an equivalent change. It is a compatible change for most normal projects, but this is a published library, so the equivalent new configuration should be `api` so that Gradle generates `compile` in the POM.

Note: you don't observe this, because the tests explicitly depend on [transitive dependencies](https://github.com/mockito/mockito-kotlin/blob/5.2.0/mockito-kotlin/build.gradle#L29), not via POM.

https://github.com/mockito/mockito-kotlin/blob/5.2.0/tests/build.gradle#L31

TWiStErRob:
* Fixes #497 

 - [x] Can you back your code up with tests? See #499

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -26,7 +26,7 @@ dependencies {
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
     compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.0.0'
 
-    implementation "org.mockito:mockito-core:5.7.0"
+    api "org.mockito:mockito-core:5.7.0"
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'com.nhaarman:expect.kt:1.0.1'


TWiStErRob:
Testing: `gradlew publishToMavenLocal` on `main` and on branch.
Diff `~\.m2\org\mockito\kotlin\mockito-kotlin\5.2.0` folder contents.
Expected: what's described in #497, actual: pass.

PR #499, opened 2023-12-02
Update integration test to prevent regression

TWiStErRob:
* Tests #498

I created a separate PR as this might be contested and iterated, and don't want to block the merging of the fix.


TWiStErRob:
This might be superseded by https://github.com/mockito/mockito-kotlin/pull/502, which a much more robust and realistic way of testing a library.

TimvdLippe:
Let's do #502 instead indeed!

PR #502, opened 2023-12-02
Move tests to included build to separate classpaths

TWiStErRob:

 - [x] Can you back your code up with tests? existing tests cover it + CI will ensure it's always working

Fixes:
```
> Configure project :tests
project ':tests' uses Kotlin 1.9.20

The Kotlin Gradle plugin was loaded multiple times in different subprojects, which is not supported and may break the build.
This might happen in subprojects that apply the Kotlin plugins with the Gradle 'plugins { ... }' DSL if they specify explicit versions, even if the versions are equal.
Please add the Kotlin plugin to the common parent project or the root project, then remove the versions in the subprojects.
If the parent project does not need the plugin, add 'apply false' to the plugin line.
See: https://docs.gradle.org/current/userguide/plugins.html#sec:subprojects_plugins_dsl
The Kotlin plugin was loaded in the following projects: ':mockito-kotlin', ':tests'
```

The solution requires some explanation while refactoring from the old one.
I'll comment on each change to clarify what it does and why it's needed.
This is the Gradle-idiomatic way of sharing code between two "independent" projects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,16 +7,22 @@ buildscript {
         classpath "org.shipkit:shipkit-changelog:1.2.0"
         classpath "org.shipkit:shipkit-auto-version:1.2.2"
         classpath "io.github.gradle-nexus:publish-plugin:1.3.0"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.20"
+        classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.9.10"
     }
 }
 
+apply plugin: 'org.gradle.base'
 apply plugin: "io.github.gradle-nexus.publish-plugin"
 apply plugin: 'org.shipkit.shipkit-auto-version'
 apply plugin: "org.shipkit.shipkit-changelog"
 apply plugin: "org.shipkit.shipkit-github-release"
 
-allprojects {
-    group = 'org.mockito.kotlin'
+def test = tasks.register("test") {
+    dependsOn gradle.includedBuild("tests").task(":test")
+}
+tasks.named("check") {
+    dependsOn test
 }
 
 tasks.named("generateChangelog") {
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -5,25 +5,14 @@ apply plugin: 'kotlin'
 apply from: '../gradle/publishing.gradle'
 apply plugin: 'org.jetbrains.dokka'
 
-buildscript {
-    ext.kotlin_version = "1.9.20"
-
-    repositories {
-        mavenCentral()
-    }
-
-    dependencies {
-        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.9.10"
-    }
-}
+group = 'org.mockito.kotlin'
 
 repositories {
     mavenCentral()
 }
 
 dependencies {
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    compileOnly "org.jetbrains.kotlin:kotlin-stdlib"
     compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.0.0'
 
     api "org.mockito:mockito-core:5.7.0"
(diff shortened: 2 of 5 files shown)


tests/build.gradle
```diff
@@ -20,14 +19,10 @@ repositories {
     mavenCentral()
 }
 
-tasks.named('compileTestKotlin') {
-    dependsOn ':mockito-kotlin:jar'
-}
```
TWiStErRob, line 25: 10. Remove this as this is automatic now. And the dependency is decoupled, not direct.

mockito-kotlin/build.gradle
```diff
@@ -5,34 +5,21 @@ apply plugin: 'kotlin'
...
-    }
-}
-
 repositories {
     mavenCentral()
 }
 
 dependencies {
-    compileOnly "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    compileOnly "org.jetbrains.kotlin:kotlin-stdlib"
```
TWiStErRob, line 13: same as 7.

build.gradle
```diff
@@ -7,9 +7,12 @@ buildscript {
         classpath "org.shipkit:shipkit-changelog:1.2.0"
         classpath "org.shipkit:shipkit-auto-version:1.2.2"
         classpath "io.github.gradle-nexus:publish-plugin:1.0.0"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.9.20"
+        classpath "org.jetbrains.dokka:dokka-gradle-plugin:1.9.10"
```
TWiStErRob, line 11: 11. Centralize `buildscript` classpath, use the same plugin versions everywhere in the project. There's only one module, so the difference is not too visible, however now there's only one place plugins are defined, other than the sample/tests project.

This is not a necessity of this refactor, but a good practice regardless. The plugin classpath is defined here, while each module/project will define what plugins are **applied** from that classpath separately. (`apply plugin:` or `plugins { id...`)

TWiStErRob: https://github.com/mockito/mockito-kotlin/pull/503 is a follow-up to this.

TWiStErRob (commented):
I numbered the comments, if you read in order it might make more sense, but no guarantees 😅

TWiStErRob:
Weird error, will rebase and try to fix.

build.gradle
```diff
@@ -7,16 +7,22 @@ buildscript {
...
 }
 
+apply plugin: 'org.gradle.base'
 apply plugin: "io.github.gradle-nexus.publish-plugin"
 apply plugin: 'org.shipkit.shipkit-auto-version'
 apply plugin: "org.shipkit.shipkit-changelog"
 apply plugin: "org.shipkit.shipkit-github-release"
 
-allprojects {
-    group = 'org.mockito.kotlin'
```
TWiStErRob, line 19: 12. moved this to `:mockito-kotlin/build.gradle` ONLY.

The CI failed with an error:
```
* What went wrong:
Could not determine the dependencies of task ':tests:test'.
> Could not resolve all task dependencies for configuration ':tests:testRuntimeClasspath'.
   > Could not resolve org.mockito.kotlin:mockito-kotlin.
     Required by:
         project :tests
      > Module version 'org.mockito.kotlin:mockito-kotlin' is not unique in composite: can be provided by [project :, project :mockito-kotlin].
```

which puzzled me because locally it was working perfectly.

I understand now what was the problem: I cloned the repo as `github-mockito-kotlin`, but the repo name is `mockito-kotlin` and `settings.gradle` does not contain a `rootProject.name` therefore the default is `mockito-kotlin`.

This above and the `allprojects` means that there are really two `org.mockito.kotlin:mockito-kotlin` artifacts, even though the rootProject (`:`) is not published.

I had the choice of setting some random `mockito-kotlin-repo` or `mockito-kotlin-root` name in `settings.gradle`, but I went with moving the `group`, because that `allprojects` is evil anyway and not necessary here as there are no other modules.

tests/build.gradle
```diff
@@ -20,15 +19,10 @@ repositories {
...
-
 dependencies {
-    implementation files("${rootProject.projectDir}/mockito-kotlin/build/libs/mockito-kotlin-${version}.jar")
+    implementation "org.mockito.kotlin:mockito-kotlin"
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation "org.mockito:mockito-core:5.7.0"
```
TWiStErRob, line 31: 13\. also removed this dependency, because it's now coming again from #498, and #499 can only be considered superseded with this line removed.

TWiStErRob:
@TimvdLippe This is ready on my side

TimvdLippe:
Thanks for all the fixes!

PR #512, opened 2024-02-22
Add support for destructured parameters in answers

Osguima3:
I think this is a useful addition as it takes benefit of a language feature instead of calling the (arguably) more verbose option of calling the `getArgument<T>(i)` method on the invocation:
```
on { intResult(any()) } doAnswer { (i: Int) -> i * 2 }
```
instead of (still supported):
```
on { intResult(any()) } doAnswer { it.getArgument<Int>(0) * 2 }
```

### DIFF
--- a/tests/src/test/kotlin/test/OngoingStubbingTest.kt
+++ b/tests/src/test/kotlin/test/OngoingStubbingTest.kt
@@ -202,6 +202,36 @@ class OngoingStubbingTest : TestBase() {
     }
 
     @Test
+    fun testOngoingStubbing_doAnswer_withDestructuredArgument() {
+        /* Given */
+        val mock = mock<Methods> {
+            on { stringResult(any()) } doAnswer { (s: String) -> "$s-result" }
+        }
+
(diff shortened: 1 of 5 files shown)


tests/src/test/kotlin/test/OngoingStubbingTest.kt
```diff
@@ -201,6 +201,20 @@ class OngoingStubbingTest : TestBase() {
         expect(result).toBe("argument-result")
     }
 
+    @Test
+    fun testOngoingStubbing_doAnswer_withDestructuredArgument() {
```
TimvdLippe, line 205: Nit: can you also add a test with 2 arguments?

TimvdLippe (commented):
Good suggestion! 1 small nit, the rest LGTM

tests/src/test/kotlin/test/OngoingStubbingTest.kt
Osguima3: Sure!

PR #521, opened 2024-07-01
Update mockito to 5.12

LarsSven:
Mockito has had 5 minor version changes since the last time this was updated. It would be nice to bump Mockito core to update the version that users of this library are using for their tests.


TimvdLippe (approved):
Thanks!

Issue #445, opened 2021-09-22
NullPointerException when mocking value class with any() matcher


yevheniisemenov:
Hi all. I guess the problem here is that I am trying to mock value class.  If I remove the `value` modifier this code works fine. Do you have any plans to add support for value classes? Also, I would be very grateful for workaround recommendations. 

Thanks!

**My dependencies:** 

```
plugins {
    kotlin("jvm") version "1.5.31"
    java
}

dependencies {
    implementation(kotlin("stdlib"))
    testImplementation("org.junit.jupiter:junit-jupiter-api:5.6.0")
    testImplementation("com.willowtreeapps.assertk:assertk-jvm:0.25")
    testImplementation("org.mockito.kotlin:mockito-kotlin:3.2.0")
    testImplementation("org.mockito:mockito-inline:3.12.4")
    testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
}
```

**Code**
```
interface MyRepository {
    fun findByEmail(email: Email): Entity
}

@JvmInline
value class Email(val email: String) {

    companion object {
        fun of(email: String): Email {
            return Email(email.lowercase())
        }
    }
}

data class Entity(val name: String)
```

**Test**
```
class MyRepositoryTest {
    @Test
    fun thisTestThrowsNpe() {
        // GIVEN
        val repository = mock<MyRepository> {
            given { it.findByEmail(any()) } willReturn { Entity("name") }
        }
    }
}
```

**The test fails with the following exception**

<details>
  <summary>Full stacktrace</summary>

```
java.lang.NullPointerException
	at org.example.MyRepositoryTest$thisTestThrowsNpe$repository$1$1.invoke(MyRepositoryTest.kt:20)
	at org.example.MyRepositoryTest$thisTestThrowsNpe$repository$1$1.invoke(MyRepositoryTest.kt:14)
	at org.mockito.kotlin.BDDMockitoKt.given(BDDMockito.kt:46)
	at org.example.MyRepositoryTest.thisTestThrowsNpe(MyRepositoryTest.kt:14)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #522, opened 2024-07-01
Added value class support

LarsSven:
This MR serves to add support for value classes to Mockito Kotlin. Value classes currently break and there is no easy way around them, many users have had issues with this: https://github.com/mockito/mockito-kotlin/issues/445

This MR adds an anyValueClass() and adds it to the any() and anyOrNull() implementation. I expect I may have done a few things non-optimally as I am not too familiar with Mockito's inner workings, but I look forward to feedback from someone who knows more about Mockito argument matchers than me to help me improve this. I do think it is important to get support for value classes.

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
@@ -50,9 +50,6 @@ inline fun <reified T : Any> any(): T {
 
 /** Matches anything, including nulls. */
 inline fun <reified T : Any> anyOrNull(): T {
-    if(T::class.isValue)
-        return anyValueClass()
-
     return ArgumentMatchers.any<T>() ?: createInstance()
 }
 
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
@@ -29,7 +29,6 @@ import org.mockito.ArgumentMatcher
 import org.mockito.ArgumentMatchers
 import org.mockito.kotlin.internal.createInstance
 import kotlin.reflect.KClass
-import kotlin.reflect.jvm.internal.impl.load.kotlin.JvmType.Object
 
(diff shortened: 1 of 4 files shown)


mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
```diff
@@ -29,6 +29,7 @@ import org.mockito.ArgumentMatcher
 import org.mockito.ArgumentMatchers
 import org.mockito.kotlin.internal.createInstance
 import kotlin.reflect.KClass
+import kotlin.reflect.jvm.internal.impl.load.kotlin.JvmType.Object
```
TimvdLippe, line 32: I don't see a usage of this class in this file. Is this import necessary and the additional dependency?

LarsSven: You were right that the import was unnecessary. I removed it.

The dependency unfortunately still is. The dependency enables reflection for Kotlin, which is necessary for detecting and introspecting value classes.

TimvdLippe:
Not sure why, but tests are failing on this PR. Can you please take a look and ensure all tests are passing?

LarsSven:
I resolved the issues with the tests, it was a relatively obscure issue that had to do with verify not being called for the test where we expected it to error out. I also went ahead and added more tests to test other scenarios like around null vs non-null and nested value classes.

vans239:
Is there any plan to add support for `eq` function?

vans239:
Maybe something like this?
```
inline fun <reified T: Any> eqValue(v: T): T {
    if (!T::class.isValue) return eq(v)
    val field = T::class.java.declaredFields.first { !Modifier.isStatic(it.modifiers) }.apply { isAccessible = true }
    val method = T::class.java.getDeclaredMethod("box-impl", field.type).apply { isAccessible = true }
    argThat(object : ArgumentMatcher<T> {
        override fun matches(argument: T) = v == (method.invoke(null, argument) as T)
        override fun toString() = ValuePrinter.print(v)
    })
    return method.invoke(null, Primitives.defaultValue(field.get(v)::class.java)) as T
}
```

PR #523, opened 2024-07-31
Add stubber (top-level) doSuspendableAnswer()

0xabadea:
This adds a top-level (stubber) `doSuspendableAnswer`  function, the counterpart of `OngoingStubbing.doSuspendableAnswer`. Rationale: while stubbing a spy I need to check that its caller has put an item into the coroutine context.

I'm not sure about the name "stubber" I used in tests. I can rename them if you come up with a better name.


TimvdLippe (approved):
Thank you!

PR #526, opened 2024-08-28
Kotlin coroutines and generic tests

hadilq:

Couldn't find Kotlin coroutines's tests and also a generic class test, so added them. Hope it's useful.

### DIFF
--- a/tests/src/test/kotlin/test/MatchersTest.kt
+++ b/tests/src/test/kotlin/test/MatchersTest.kt
@@ -49,7 +49,7 @@ class MatchersTest : TestBase() {
         mock<Methods>().apply {
             runTest {
                 coroutinesClosed(Closed())
-                //TODO verify(this).coroutinesClosed(any())
+                verify(this@apply).coroutinesClosed(any())
             }
         }
     }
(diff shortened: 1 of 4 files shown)


tests/src/test/kotlin/test/MatchersTest.kt
```diff
@@ -35,6 +36,24 @@ class MatchersTest : TestBase() {
...
+            verify(this).classClosed(any())
+        }
+    }
+
+    @Test
+    fun anyCoroutinesClosedClass() {
+        mock<Methods>().apply {
+            runTest {
+                coroutinesClosed(Closed())
+                //TODO verify(this).coroutinesClosed(any())
```
TimvdLippe, line 52: Does this mean this currently doesn't work as expected? Do you want to submit a fix as well for this?

TimvdLippe (approved):
Tests look good, thanks! 1 question about the TODO.

tests/src/test/kotlin/test/MatchersTest.kt
TimvdLippe: Let's uncomment the code here and add a `@Disabled` to this test with an explanation. Then we can fix that in the next PR.

hadilq: interestingly, I removed the TODO and it passed the test!! I cannot remember what I have done before :D

TimvdLippe (approved):
Thanks!

PR #534, opened 2025-04-10
Update mockito

LarsSven:
It's been some time since Mockito has been upgraded, so this upgrades Mockito to the newest version.

Maybe it would also make sense if this is accepted to upgrade Kotlin to some K2 version? Since a relatively large part of the ecosystem has moved to K2 I believe.


TimvdLippe (approved):
Thanks! Feel free to update to K2. We probably need to make that a breaking change.

PR #535, opened 2025-04-14
Upgrade Kotlin and Gradle

LarsSven:
This is a followup of https://github.com/mockito/mockito-kotlin/pull/534#pullrequestreview-2756992284

It upgrades Kotlin and Gradle to a newer version.

These are some pretty major upgrades so I do recommend also locally testing it before approving/merging.


LarsSven:
The current CI failure seems to be a random failure that probably needs to be rerun?

LarsSven:
The `doReturn_withGenericIntReturnType_on` test specifically fails on the upgrade from Kotlin 1.9 to 2.0 (2.0.0 is the first version that starts failing). It is not fully clear to me why it starts failing, especially because while `on` does not work, `onGeneric` does work fine, and the generated error message by mockito-kotlin also suggests that you may need to use onGeneric in this case.

I am not too deep in this codebase to understand what `on` and `onGeneric` do specifically and why one would fail with an NPE and the other wouldn't.

LarsSven:
@TimvdLippe would you perhaps be able to take a look at this?

TimvdLippe:
Unfortunately I am not versed in Kotlin, so I can't assist with investigating Kotlin specific failures. Maybe somebody else with more knowledge of the Kotlin compiler can help out? It's not the first time Kotlin does something that is unusual compared to the other JVM languages.

Gungy2:
> I had another look at it but it almost feel like maybe the test is not correct. Why did `on` work on generic methods to begin with, aren't you supposed to always use `onGeneric` for generic methods?

I think I agree with this assessment, especially because the `doReturn_withGenericIntReturnType_onGeneric` method works correctly. Probably the way forward is just deleting that test?

LarsSven:
@TimvdLippe I removed the broken test

TimvdLippe (approved):
Thanks all for the assistance! I will publish a new version of `mockito-kotlin` with this change.

PR #554, opened 2025-11-29
Enhance tests covering various facets of stubbing, this is groundwork towards implementing improved suspend function support

m-koops:
As discussed in #553, that PR had become very large and hard to review, due to its many changes all combined in 1 squashed commit. This PR is a first step to break down that big PR and deliver a train of smaller PRs that ultimately lead to a combined result comparable with #553. PR #553  is a proper reference to describe the context and aim of the changes, but in the end I expect #553 not to be merged as it will get superceeded by the train of smaller PRs.

This PR is taking all enhancements to the test code that can be done without touching any class of the current mainline code.
The commits in this PR are not (yet) squashed, as each of the commits represent a logic step in the evolution of the test code. Each commit is a valid and complete stage in the code history, as each commit has a proper commit message and a succesfull Gradle build.

I would advice to review the PR commit by commit, as each commit on its own is small and should be well comprehensible.  

Merging this PR would set the stage to introduce the enhancements to the mainline code that I will present in smaller and more manageable PRs as well.

### DIFF
--- a/tests/src/test/kotlin/test/Classes.kt
+++ b/tests/src/test/kotlin/test/Classes.kt
@@ -98,6 +98,9 @@ interface SynchronousFunctions {
     fun valueClass(v: ValueClass)
     fun nullableValueClass(v: ValueClass?)
     fun nestedValueClass(v: NestedValueClass)
+    fun valueClassResult(): ValueClass
+    fun nullableValueClassResult(): ValueClass?
+    fun nestedValueClassResult(): NestedValueClass
 }
 
 interface SuspendFunctions {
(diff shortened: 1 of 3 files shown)


TimvdLippe (approved):
Reviewing per commit was crucial to understand all the moving pieces. Thank you for taking your time doing so and all the test suite improvements. Given that no main source code is changed and only tests were updated, I trust the process that these changes are fine.

Given that all commits themselves also pass (nice!), I will rebase and merge to preserve the well-curated history. For PRs that touch main source code, I want to squash these, so make sure they are smaller than this one.

Issue #555, opened 2025-12-03
Issues stubbing and capturing value classes in latest snapshot


jselbo:
Hello,
I think there are issues introduced by the recent commits
d8fe7c78ce45cba648b9a1977b7da9993ac05005 "Add value class support to KArgumentCaptor"
cd674c44188fb2d7ee018fab7467498991e0b2a0 "Add value class support to eq() argument matcher"

I am trying to integrate the latest mockito-kotlin snapshot build with our suite of tests and I'm encountering these issues with value classes that wrap primitive types (like `kotlin.time.Duration`):

1. `anyOrNull` with nullable primitive value class matcher throws IllegalStateException

See 7a19a0a2863e118d46a6926d8351dc8298e2045a for a repro and a fix - I can open a PR if desired but I don't want to step on the toes of existing work

2. `any`/`eq` don't work for primitive value class

```
@JvmInline value class PrimitiveValueClass(val value: Long)

// method:
fun nullablePrimitiveValueClass(v: PrimitiveValueClass?)

// test:
mock<SynchronousFunctions>().apply {
  nullablePrimitiveValueClass(PrimitiveValueClass(123))
  verify(this).nullablePrimitiveValueClass(any())
}
```
fails with:
```
Argument(s) are different! Wanted:
synchronousFunctions.nullablePrimitiveValueClass-2LcYgm0(
    <any long>
);
-> at test.MatchersTest$AnyOrNullMatchersTest.anyOrNullNullableValueClass(MatchersTest.kt:375)
Actual invocations have different arguments:
synchronousFunctions.nullablePrimitiveValueClass-2LcYgm0(
    PrimitiveValueClass(value=123)
);
```

I think the `any()` wrongly unwraps the value class type when it is not actually inlined here.

I also saw some failures related to `kotlin.Result<T>` unwrapping but I'm not sure if that's a separate issue.

cc: @m-koops

PR #557, opened 2025-12-08
Fix anyOrNull matcher with value class wrapping primitive type

jselbo:
Fixes the first problem mentioned in #555
"anyOrNull with nullable primitive value class matcher throws IllegalStateException"

`createInstance` value class logic assumed the value class can be constructed with a `null` value, but this doesn't work when the wrapped type is a primitive.

### DIFF
--- a/tests/src/test/kotlin/test/ArgumentCaptorTest.kt
+++ b/tests/src/test/kotlin/test/ArgumentCaptorTest.kt
@@ -2,6 +2,7 @@ package test
 
 import com.nhaarman.expect.expect
 import com.nhaarman.expect.expectErrorWithMessage
+import org.junit.Ignore
 import org.junit.Test
 import org.mockito.kotlin.*
 import java.util.*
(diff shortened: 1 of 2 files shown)


m-koops:
Hi Joshua, many thanks for providing this PR. 

The extra tests that you have added, but which you had to disable with @Ignore as they fail at this moment, shed an interesting light on how Kotlin hands over value classes to Java code.  So, apart from your direct code improvement, #555 and this PR have provided valueable additional input on the weak spots of value class handling in Mockito-Kotlin.

So far, my impression was that value class are always unboxed when handed over to Java code. But the new test cases proof that in case of primitives as inner value, the value classes are handed over to the Mockito Java code as-is, say boxed.

@TimvdLippe, apart from the unused import that I marked with a comment, this PR is ready to be accepted. 

As follow-up, I will have a look into the boxing and unboxing behavior of value classes in general and hopefully I can soon provide a fix that will make the new ignored test cases succeed. 

Cheers, Mark

PR #560, opened 2025-12-10
Update MockSettings usage

jselbo:
- Remove `@Incubating` from `useConstructor` and `outerInstance` params. mockito-core removed `@Incubating` from these methods on `MockSettings` in the 4.0.0 release: https://github.com/mockito/mockito/commit/7032574ce1fb65e5dbdcf24f65f87766d7124db3
- ~~Remove `lenient` param since it is deprecated in `MockitoSettings` in mockito-core in favor of `strictness`~~
- Deprecate `lenient` param since it is deprecated in `MockitoSettings` in mockito-core
- Add new `strictness` param
- Add support for `defaultAnswer` setting for `mockStatic` API

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
@@ -50,6 +50,7 @@ import kotlin.reflect.KClass
  * @param stubOnly A stub-only mock does not record method invocations, thus saving memory but disallowing verification of invocations.
  * @param useConstructor Mockito attempts to use constructor when creating instance of the mock.
  * @param outerInstance Makes it possible to mock non-static inner classes in conjunction with [useConstructor].
+ * @param lenient (DEPRECATED) Lenient mocks bypass "strict stubbing" validation.
  * @param strictness Specifies strictness level for the mock.
  */
 inline fun <reified T : Any> mock(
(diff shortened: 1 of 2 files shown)


jselbo:
Yeah. Let me break each of these into separate PRs.

TimvdLippe:
We can also keep the lenient boolean, mark it as deprecated and then convert it to the correct strictness value. And if somebody supplies both, the strictness value takes precedence.

jselbo:
Unfortunately Kotlin's `@Deprecated` annotation cannot be used on parameters, only on the entire method.

I'm not keen on introducing another `mock` overload, though that seems to be the only way to avoid the breaking change. What do you think?

TimvdLippe:
I am confused, since I thought that we wouldn't need a `mock` overload given that they are all nullable? E.g. why wouldn't this work:

```kt
inline fun <reified T : Any> mock(
    extraInterfaces: Array<out KClass<out Any>>? = null,
    name: String? = null,
    spiedInstance: Any? = null,
    defaultAnswer: Answer<Any>? = null,
    serializable: Boolean = false,
    serializableMode: SerializableMode? = null,
    verboseLogging: Boolean = false,
    invocationListeners: Array<InvocationListener>? = null,
    stubOnly: Boolean = false,
    useConstructor: UseConstructor? = null,
    outerInstance: Any? = null,
    lenient: Boolean = false
    strictness: Strictness? = null,
): T {
```

As I understand it, this is backwards compatible as well, since we only add a new parameter and we keep the existing ones.

Sure, we can't mark the one parameter as deprecated, but we can denote that in the method description. And we can add that to a migration guide when we a publish a new major version.

TimvdLippe (approved):
Nice thanks!

PR #561, opened 2025-12-12
Follow-up on primitive value classes

m-koops:
This PR is a direct follow-up to #557 , in which @jselbo introduced some test cases that proved Mockito-Kotlin to fail on handling (nullable) primitive value types in eq matcher and argument captor.

Some background: Kotlin compiler will typically pass on value classes unboxed to Java calls. While unboxing, it will apply the nullability of the value class to the unboxed inner type. But with one exception: when the inner type is a Java primitive then nullability cannot be applied to that primitive. In that case Kotlin compiler passes on the unboxed value type into Java calls.
This exceptional behavior is what is breaking MOckito-Kotlin.

To fully cover handling of (nullable) primitive value class types, tests have been added/activated for those types in eq and any matchers, in argument captors and as return values in synchronous and suspendable functions.

Additionally, the boxing and unboxing logic has been extracted to a new class ValueClassSupport, to facilitate for reuse of the logic and to test the logic with dedicated unit tests.

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/ArgumentCaptor.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/ArgumentCaptor.kt
@@ -36,13 +36,40 @@ import kotlin.reflect.typeOf
 
 /**
  * Creates a [KArgumentCaptor] for given type.
+ *
+ * Caution: this factory method cannot be used to create a captor for a suspend
+ * function, please refer to [suspendFunctionArgumentCaptor] for that.
+ * This incompatibility is caused by the use of `typeOf<T>()` which is the way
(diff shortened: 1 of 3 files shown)


m-koops:
I would like to suggest to first focus on getting this PR merged, before merging #556. After merging this one, and rebasing #556, I can adjust documentation of new API to that of the other API. Also the added test cases should be updated to align with the other tests. I can do so in an additional commit after rebasing.

m-koops:
> Can you comment on KT-47562 linking to this PR and saying this complicates the public API for a very popular mocking framework? Maybe a bump will motivate Jetbrains to fix the issue with `typeOf()`

I've just posted a comment explaining the troubles encountered here due to `typeOf()` being incompatable with suspend functions.

Issue #533, opened 2025-02-03
Cannot mock boxed types


wegendt-bosch:
This test fails:

```kotlin
@ExtendWith(MockitoExtension::class)
class MockitoTest {
  class Consumer<T>(val fn: (T) -> Unit)
  class Mocked {
    fun <T> callConsumer(consumer: Consumer<T>, value: T) {
      consumer.fn(value)
    }
  }

  @Mock
  private lateinit var mocked: Mocked

  @Test
  fun testPrimitives() {
    var functionCalled = false
    val aConsumer = Consumer<UShort> {}
    whenever(mocked.callConsumer(same(aConsumer), any())).then { functionCalled = true }

    mocked.callConsumer(aConsumer, 42u)
    assertTrue(functionCalled)
  }
}
```

but works when you change the type parameter to `Short`:

```kotlin
@ExtendWith(MockitoExtension::class)
class MockitoTest {
  class Consumer<T>(val fn: (T) -> Unit)
  class Mocked {
    fun <T> callConsumer(consumer: Consumer<T>, value: T) {
      consumer.fn(value)
    }
  }

  @Mock
  private lateinit var mocked: Mocked

  @Test
  fun testPrimitives() {
    var functionCalled = false
    val aConsumer = Consumer<Short> {}
    whenever(mocked.callConsumer(same(aConsumer), any())).then { functionCalled = true }

    mocked.callConsumer(aConsumer, 42)
    assertTrue(functionCalled)
  }
}
```

m-koops:
PR #554 was mistakingly referencing to this issue, as I was aiming to reference PR #553. But the good news: the changes I'm working on to improve value class support in these PRs will indirectly also help to fix this issue as unsigned integers are a subclass of value classes in Kotlin.

Issue #363, opened 2019-09-29
[Feature request] Support unsigned numbers


neworld:
Info:
- Reference: https://kotlinlang.org/docs/reference/basic-types.html#unsigned-integers
- KEEP: https://github.com/Kotlin/KEEP/blob/master/proposals/unsigned-types.md

Test case:
```
    @Test
    fun mockitoTest() {
        val fixture: UnsignedFixture = mock()

        fixture.ubyte(42u)
        fixture.ushort(42u)
        fixture.uInt(42u)
        fixture.uLong(42u)

        verify(fixture).ubyte(eq(42u))
        verify(fixture).ushort(eq(42u))
        verify(fixture).uInt(eq(42u))
        verify(fixture).uLong(eq(42u))
    }

    interface UnsignedFixture {
        fun ubyte(value: UByte)
        fun ushort(value: UShort)
        fun uInt(value: UInt)
        fun uLong(value: ULong)
    }
```

Error:
```
Argument(s) are different! Wanted:
unsignedFixture.uInt-WZ4Q5Ns((UInt) 42);
-> at lt.libredrop.peerdiscovery.PeerDiscoverySpecsComplainTest.mockitoTest(PeerDiscoverySpecsComplainTest.kt:72)
Actual invocations have different arguments:
unsignedFixture.uInt-WZ4Q5Ns((Integer) 42);
-> at lt.libredrop.peerdiscovery.PeerDiscoverySpecsComplainTest.mockitoTest(PeerDiscoverySpecsComplainTest.kt:67)
```

bohsen:
Duplicate of #309 

@neworld The issue mentions a workaround, but not sure it fits your needs. Also as long as Mockito-kotlin depends on mockito (java) internally I don't think there's an easy solution to this.

PR #569, opened 2025-12-20
Improving stubbing support for suspend functions and value classes

m-koops:
This PR provides a fix for #456 and will prevent future confusion of users around the stubbing of suspend functions.
This PR also does the groundwork for #533 and #363  (support for unsigned integers), as the unsigned integers are a flavor of value classes within Kotlin.

This PR will conclude the sequence of PRs that were the result of breaking down the massive PR #553 into multiple smaller changesets.

=====

This PR further improves stubbing support for suspend functions/coroutines in combination with (potentially nullable and /or primitive) value classes. 

In the various Mockito-kotlin stubbing functions the Mockito answers are now all wrapped in the new `CoroutineAwareAnswer`, which allows the answers to lazily do the right amount of boxing and specific handling, right at the moment that the answer is evaluated by Mockito Core. 

On evaluation (by a call of the `answer()` function) `CoroutineAwareAnswer` will first detect if the invocation, it acts on, is tied to a suspend function. If so, the `CoroutineAwareAnswer` will take care of the following:
 - it will wrap the original answer in a `SuspendableAnswer` (an existing class) to ensure that the result of the answer is wired properly into the continuation callback infrastructure of the invoked suspend function
- it will enforce suspension within the `SuspendableAnswer` execution to ensure proper wiring to the callback function
- lastly, it will box results yielding from real method calls in the mock to a value class, if needed so

In case the `CoroutineAwareAnswer` is being tied to a synchronous method/function call it will just forward its invocation to the nested delegate answer (the original Mockito answer). In this scenario the wrapping of the answer in a `CoroutineAwareAnswer` should be without effect.

As long as the user is applying the Mockito-kotlin stubbing functions, this library should now take care of dealing with all intricacies related to the combinations of suspend functions, (primitive) value classes and nullability. As such, the user does not have to wrap a value class return value in a rather 'magic' lamda anymore, like this:
```kotlin
doSuspendableReturn {
    delay(1)
    valueClass
}
```
Instead the user can now specify the value class return value of a suspend function call by simply applying: 
```
doReturn(valueClass)
``` 

The many existing unit tests prove that overall this new implementation does deliver the same result as before.
Some tests in `CoroutinesOngoingStubbingTest` could be simplified proving that `doReturn()` now works fine for suspend functions with value class retrun values.

Additionally the functions `doCallRealMethod()` and `callRealMethod()` have been added to OngoingStubbing, and tests for sync and suspend functions have been added to test them.


m-koops:
@jselbo Can you please run your test suits against this new (lazy) implementation? I aim for no regressions at all.

jselbo:
I checked one failure and it looks like an ordering issue where a test (the body wrapped in `runTest(StandardTestDispatcher()) { .. }` and it also overrides the main dispatcher) - invokes a function that launches a coroutine like `CoroutineScope(Dispatchers.IO).launch { ... }`. With this patch, the coroutine does not execute immediately though we expect it to be.

I suspected the `delay(1)` in `SuspendableAnswer` and sure enough if I comment it out the test succeeds. Is there a way to make this work without the delay? I don't fully understand the interaction here but it seems to have some unintended effects on scheduling.

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/CoroutineAwareAnswer.kt
```diff
@@ -0,0 +1,108 @@
...
+    @Suppress("UNCHECKED_CAST")
+    private class SuspendableAnswer(private val block: suspend (KInvocationOnMock) -> Any?) :
+        Answer<Any?> {
+        override fun answer(invocation: InvocationOnMock): Any? {
+            val unboxNonPrimitiveValueClasses: suspend (KInvocationOnMock) -> Any? =
+                { invocationOnMock ->
+                    block.invoke(invocationOnMock)?.let {
+                        if (it::class.isValue) {
+                            val unboxed = it.unboxValueClass()
+                            if (unboxed is Number) it else unboxed
```
jselbo: What about boolean? Wouldn't this be more appropriate? `if (unboxed::class.java.isPrimitive)`

TimvdLippe (approved):
Thank you Joshua for the review and verification on the test suite!

jselbo:
cc @m-koops  sorry to tag you on an old PR but I am trying to improve the wiki pages. I added https://github.com/mockito/mockito-kotlin/wiki/Coroutines-%5BWork%E2%80%90in%E2%80%90Progress%5D to cover coroutine support. 

What's not totally clear to me is why we have all these special APIs to deal with suspendable functions like `verifyBlocking`, `whenever {}`, ...  I can see this being confusing to users, and even I'm confused.

Tests using coroutines will likely be using https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-test/ and wrapping the test block with `runTest {}`. In a coroutine context block, you can just call suspendable functions normally. 

Is there a reason tests for code using coroutines would not be using a `runTest` or similar?

Also, it would be great if you could update that wiki page to cover anything I missed 😄

m-koops:
Hi Joshua,


Thank you very much for your initiative.

Traveling currently without laptop I cannot check details in the code.
Overall I can say that, even in the scope of `runtest` suspend functions
need special boxing/unboxing handling when stubbing mocks.

In a first iteration some 'special' mockito api like onBlocking and
verifyBlocking have been introduced. This required users to be aware of
these special case api.

My changes in 6.2.0 were aiming to arrange the special handling 'under the
hood' with additional overload functions.  The 'special' blocking functions
are now marked as deprecated and are only left in the code for backwards
compatibility.

I was aiming to provide 1 unified api that deals well with both blocking
and coroutine function calls, e.g. the `on`, `whenever` and `verify`
functions with its overloads.

When back home, I'm happy to help you to improve the wiki documentation.

Issue #570, opened 2026-01-14
6.2.0 regression stubbing broken for non-coroutines project


vladimirfx:
In Spring Boot project without coroutines runtime dependencies:

```
kotlinx.coroutines.BuildersKt
java.lang.ClassNotFoundException: kotlinx.coroutines.BuildersKt
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:580)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:490)
	at org.mockito.kotlin.OngoingStubbingKt.whenever(OngoingStubbing.kt:96)
	at org.mockito.kotlin.KStubbing.on(KStubbing.kt:108)
	at Test.mockFindRelevantPlaceExtended(PlaceHandlerTest.kt:2133)
	at Test.check data for Estados Unidos(PlaceHandlerTest.kt:1636)

```

Please provide non suspend alternatives for stubbing API.

TimvdLippe:
Fixed in #575 and will publish 6.2.1 now.

PR #571, opened 2026-01-14
Make kotlinx-coroutines-core (with runBlocking in it) available as transitive dependency.

m-koops:
Fixes issue #570 

Due to the code changes runBlocking is needed in every use of Mockito-kotlin, also for projects without coroutines.
That's why Mockito-kotlin should declare 'kotlinx-coroutines-core' as a transitive runtime dependency.

This PR is superseded by PR #575

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -16,7 +16,7 @@ repositories {
 
 dependencies {
     compileOnly "org.jetbrains.kotlin:kotlin-stdlib"
-    compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2'
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2'
     implementation "org.jetbrains.kotlin:kotlin-reflect:2.1.20"
 
     api "org.mockito:mockito-core:5.20.0"


m-koops:
> 2. I agree that we should keep this optional if possible. Can we update the code to handle the case where classes don't exist? We use that pattern in `mockito-core` as well: https://github.com/mockito/mockito/blob/d16fcfc274d7ca03a2b4bdc22dd7c3ec6dac8690/mockito-core/src/main/java/org/mockito/internal/junit/ExceptionFactory.java#L23

This have looked in a solution like this myself. But the issue is that the argument of the whenever function is now marked as suspend function/lambda. One can pass either a regular lambda or a suspend function into this function. But once marked suspending, there is no way to unmark it as suspending. And the passed in function therefor can only be invoked in a coroutine scope. Hence `runBlocking` is needed.

It's a pity that the Kotlin compiler does not yet allow to specify 2 overloads of the whenever function, 1 for regular lambdas, the other for suspend lambdas. Once that option is provided by the compiler it is way easier to support the non-coroutine without the need for a dependency to kotlinx-coroutines-core. 

Tomorrow morning I will dive further into the 2 regressions reported. And I will do my best to come with the least intrusive fixes to them.

PR #578, opened 2026-01-19
Fix eq() matcher NPE with nullable value class and null value

### LINKED ISSUE #577: `eq()` matcher throws NPE when used with nullable value class and null value
## Description

After upgrading from 6.1.0 to 6.2.0, `eq()` matcher throws `NullPointerException` when:
- The parameter type is a **nullable value class** (e.g., `MyValueClass?`)
- The actual value is **`null`**

This is a regression introduced in #545.

## Reproduction

```kotlin
@JvmInline
value class MyValueClass(val value: Long)

interface Foo {
    fun bar(x: MyValueClass?)
}

@Test
fun `eq with nullable value class and null value`() {
    val foo = mock<Foo>()
    foo.bar(null)
    
    verify(foo).bar(eq(null))  // Throws NPE in 6.2.0+
}
```

## Error

```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "value" is null
    at org.mockito.kotlin.MatchersKt.eqValueClass(Matchers.kt:91)
    at org.mockito.kotlin.MatchersKt.eq(Matchers.kt:40)
```

## Affected Versions

| Version | Status |
|---------|--------|
| 6.1.0   | ✅ Works |
| 6.2.0   | ❌ Broken |
| 6.2.1   | ❌ Broken |

## Root Cause

In `Matchers.kt`, the `eq()` function checks `T::class.isValue` before checking if `value` is null:

```kotlin
inline fun <reified T : Any?> eq(value: T): T {
    if (T::class.isValue) return eqValueClass(value)  // Enters this branch even when value is null
    // ...
}
```

When `T = MyValueClass?` and `value = null`:
1. `T::class.isValue` returns `true` (compile-time type is a value class)
2. `eqValueClass(null)` is called
3. `require(value::class.isValue)` throws NPE because `null::class` is invalid

KonaEspresso94:

 - [x] Please run `./gradlew spotlessApply :tests:spotlessApply` for auto-formatting.

---

## Summary

This PR fixes the NPE issue in `eq()` matcher when used with nullable value class and null value.

Fixes #577

## Problem

After #545 added value class support to `eq()`, using `eq(null)` with a nullable value class type throws `NullPointerException`:

```kotlin
@JvmInline
value class MyValueClass(val value: Long)

interface Foo {
    fun bar(x: MyValueClass?)
}

// This throws NPE in 6.2.0+
verify(foo).bar(eq(null))
```

## Root Cause

In `eqValueClass()`, the code accesses `value::class` before checking if `value` is null:

```kotlin
inline fun <reified T> eqValueClass(value: T): T {
    require(value::class.isValue) { ... }  // NPE when value is null
    ...
}
```

## Solution

Add a null check at the beginning of `eqValueClass()`:

```kotlin
if (value == null) {
    @Suppress("UNCHECKED_CAST")
    return ArgumentMatchers.eq<T>(null) as T
}
```

This follows the existing pattern used in `toKotlinType()` in `ValueClassSupport.kt`.

## Changes

- `Matchers.kt`: Add null check in `eqValueClass()`
- `MatchersTest.kt`: Add test cases for nullable value class with null value

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Matchers.kt
@@ -88,9 +88,14 @@ inline fun <reified T> anyValueClass(): T {
 
 /** Matches an argument that is equal to the given Kotlin value class value. */
 inline fun <reified T> eqValueClass(value: T): T {
+    if (value == null) {
+        @Suppress("UNCHECKED_CAST")
+        return ArgumentMatchers.eq<T>(null) as T
+    }
+
     require(value::class.isValue) { "${value::class.qualifiedName} is not a value class." }
 
-    val unboxed = value?.unboxValueClass()
+    val unboxed = value.unboxValueClass()
     val matcher = AdditionalMatchers.or(ArgumentMatchers.eq(value), ArgumentMatchers.eq(unboxed))
 
     return (matcher ?: unboxed).toKotlinType(T::class)
--- a/tests/src/test/kotlin/test/MatchersTest.kt
+++ b/tests/src/test/kotlin/test/MatchersTest.kt
@@ -672,6 +672,22 @@ class MatchersTest : TestBase() {
                 verify(this).nestedValueClass(eq(nestedValueClass))
             }
         }
+
+        @Test
+        fun eqNullableValueClass_nullValue() {
+            mock<SynchronousFunctions>().apply {
+                nullableValueClass(null)
+                verify(this).nullableValueClass(eq(null))
+            }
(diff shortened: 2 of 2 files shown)


m-koops:
@KonaEspresso94 Thank you very much for raising the NPE issue and providing your fix in this PR. While reviewing the change, I had to make some slight adjustments.

Therefore I decided to prepare a new commit/PR (#579). @TimvdLippe can you review that new PR?

KonaEspresso94:
@TimvdLippe @m-koops Thanks for your quick feedback. I’ll close this PR.

PR #586, opened 2026-03-05
Add helper to mock module-level extension functions

jselbo:
Kotlin lang ref: https://kotlinlang.org/docs/extensions.html

New syntax sugar feature: a helper to mock non-member extension functions.
Note the implementation is just `mockStatic` on the declaring `FooKt` class, so all declared extension functions would be mocked at the same time. There's no way to mock only one of the extension funs.

Also update the spotless task so you don't need to include `:tests:spotlessApply`

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
@@ -28,6 +28,9 @@ package org.mockito.kotlin
 import java.lang.reflect.Modifier
 import kotlin.DeprecationLevel.ERROR
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
+import kotlin.reflect.full.extensionReceiverParameter
+import kotlin.reflect.jvm.javaMethod
 import org.mockito.MockSettings
 import org.mockito.MockedConstruction
 import org.mockito.MockedStatic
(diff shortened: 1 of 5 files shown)


mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
```diff
@@ -230,6 +233,51 @@ inline fun <reified T> mockConstruction(
...
+ * ```
+ *
+ * For overloaded extension functions, specify the type to disambiguate:
+ * ```
+ * fun String.isHello(): Boolean = this == "Hello"
+ * fun String.isHello(mood: String): Boolean = this == "Hello" && mood == "happy"
+ *
+ * val ref: KFunction2<String, String, Boolean> = String::isHello
+ * mockExtensionFun(ref).use {
+ *     whenever("test".isHello("sad")).thenReturn(true)
```
hick209: I know this is like a static method extension as that's what those basically are for the JVM.
My question here is, does here mean that the input params, including the receiver, have to match?
In other words, does `"anotherTest".isHello("sad")` return `false`?

If so, I wonder if we could come up with an elegant way to do something like `whenever(any<String>.isHello(mood)).thenReturn(true)`

hick209 (changes requested):
Overall this LGTM.
I just have one clarifying question.

I think we should update the kdoc and tests to make that clear for other people as well, which is the sole reason I request changes here.

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/Mocking.kt
jselbo: Great question. this is equivalent to doing
```
whenever(ref.invoke("test", "sad")).thenReturn(true)
```
so I tested and surprisingly yes we can use matchers in just the same way, even with the regular function invoke syntax.

```
whenever(any<String>().isHello(eq("sad")).thenReturn(true)
```
which reads kind of weird.

But you still have to follow the rule of all args being matchers. I think for that reason, we should encourage usage of this API with the explicit `f.invoke` because it makes the stubbing more like traditional stubbing syntax. I will update the docs and tests to reflect this.

jselbo: Actually, the `.invoke()` syntax with matchers won't work because the compiled code for `KFunction.invoke` includes null checks that matchers can't satisfy unfortunately.

hick209 (approved):

raphw (approved):
Clean and well documented change. Looks good to me.

PR #587, opened 2026-03-11
Add first-class support for mocking object singletons

### LINKED ISSUE #536: How to properly mock companion object methods?
Consider a Kotlin class with companion object:
```
class MyClass {
    companion object {
        fun foo(): String {
            return "unmocked"
        }
    }
}
```

Intuitively, we might expect to use `mockStatic(MyClass.Companion::class.java)` and stub `foo()`. But this doesn't work as expected. (Note: if you mark with the method with `@JvmStatic`, you can stub the generated static wrapper method but only Java callsites will invoke this.)

We found a way to make stubbing companion object methods work by using `mockConstruction`:
```
mockConstruction(MyClass.Companion::class.java).use {
    whenever(MyClass.foo()).thenReturn("Mocked")

    val result = MyClass.foo()
    assertEquals("Mocked", result)
    verify(MyClass).foo()
}
```
This works because `foo()` as compiled is actually an instance method on the inner class `MyClass$Companion`. A callsite invoking the method is equivalent to:
```
val myClassCompanionInstance = MyClass.Companion
myClassCompanionInstance.foo()
```
where MyClass bytecode is equivalent to (expressed in Java):
```
class MyClass {
  static final Companion Companion;
  static {
    Companion = new Companion();
  }
  ....
}
```

Meanwhile, StackOverflow answers recommend refactoring the code to have a wrapper method, or use competing mocking frameworks like MockK or PowerMock:
 https://stackoverflow.com/questions/53174839/unit-testing-verifying-a-companion-object-method-is-called-mocking-a-companion

Is using `mockConstruction` really the intended way to stub methods on Kotlin companion objects? If so, can you add this guidance to the wiki?

jselbo:
Fixes: #536

This builds off the new `mockSingleton` feature of Mockito-core which enables thread-local intercepting instance methods of existing objects. This was added to enable mocking Kotlin `object` and `companion object` singletons. See https://github.com/mockito/mockito/pull/3762 

Very excited to finally add this feature!

This is comparable to MockK's `mockkObject`.

### DIFF
--- a/mockito-kotlin/build.gradle
+++ b/mockito-kotlin/build.gradle
@@ -19,7 +19,7 @@ dependencies {
     compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2'
     implementation "org.jetbrains.kotlin:kotlin-reflect:2.1.20"
 
-    api "org.mockito:mockito-core:5.20.0"
+    api "org.mockito:mockito-core:5.23.0"
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'com.nhaarman:expect.kt:1.0.1'
--- a/tests/src/test/kotlin/test/MockingTest.kt
+++ b/tests/src/test/kotlin/test/MockingTest.kt
@@ -487,12 +487,29 @@ class MockingTest : TestBase() {
     }
 
     @Test
-    fun mockStatic_Object_With_JvmStatic_method() {
-        mockStatic<MyObject>().use { mock ->
-            mock.whenever { MyObject.helloStatic() }.thenReturn("stubbed")
+    fun mockObject_JvmStatic_method() {
+        mockObject(MyObject).use {
+            whenever(MyObject.helloStatic()).thenReturn("stubbed_via_mockObject")
 
-            expect(MyObject.helloStatic()).toBe("stubbed")
+            expect(MyObject.helloStatic()).toBe("stubbed_via_mockObject")
         }
+    }
+
+    @Test
+    fun mockObject_mixed_regular_and_JvmStatic_methods() {
+        mockObject(MyObject).use {
+            whenever(MyObject.hello()).thenReturn("regular_stubbed")
+            whenever(MyObject.helloStatic()).thenReturn("static_stubbed")
+
+            expect(MyObject.hello()).toBe("regular_stubbed")
+            expect(MyObject.helloStatic()).toBe("static_stubbed")
+            expect(it.isMockingStatic).toBe(true)
+        }
+    }
+
+    @Test
+    fun mockObject_without_JvmStatic_does_not_use_mockStatic() {
+        mockObject(MyObjectNoStatic).use { expect(it.isMockingStatic).toBe(false) }
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


tests/src/test/kotlin/test/MockingTest.kt
```diff
@@ -470,6 +471,86 @@ class MockingTest : TestBase() {
...
+                mockObject(m)
+            }
+    }
+
+    object MyObject {
+        fun hello(): String = "unstubbed"
+
+        // This will compile to a static method callable by both Java and Kotlin.
+        // It must be mocked and stubbed with mockStatic!
+        @JvmStatic fun helloStatic(): String = "static_unstubbed"
```
jselbo, line 560: This is the only gotcha - `@JvmStatic` marked methods in `object` (but not companion object) compile only to a static method and must be mocked with `mockStatic`.

I plan to update the wiki page with more info about mocking objects after this is merged!

mockito-kotlin/build.gradle
```diff
@@ -19,7 +19,7 @@ dependencies {
     compileOnly 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.2'
     implementation "org.jetbrains.kotlin:kotlin-reflect:2.1.20"
 
-    api "org.mockito:mockito-core:5.20.0"
+    api "org.mockito:mockito-core:5.23.0"
```
hick209, line 22: Oh, wow! I'm surprised this is not extracted to a param somewhere so we'd have a single source of truth for the version across the mockito libs.
Interesting to know, but certainly not something related to this PR.

tests/src/test/kotlin/test/MockingTest.kt
hick209: Dang, this is super tricky!
I wonder if we could signal this somehow to people.

I know this is also what libs like MockK do, but if we could surface that somehow, then it could help Mockito stand out. That'd be a whole new discussion though, outside this PR.

## My random idea
What if Mockito had a lint lib that people could easily pull and use?

hick209 (approved):

mockito-kotlin/build.gradle
jselbo: This is the only place we specify the mockito-core version in this repo. But there duplicated versions across the build.gradle files. As I follow-up I can convert this to use Gradle version catalog

tests/src/test/kotlin/test/MockingTest.kt
jselbo: Lint is an interesting idea, though it would be hard to get traction. I'd prefer a built-in solution.

I also hate that this gotcha exists. We could try to be smart and invoke `mockStatic` under the hood if we detect `@JvmStatic` methods. Let me explore this option.

hick209: Wow! I LOVED your solution! Really clever, ship it! 🚀

hick209 (approved):

mockito-kotlin/build.gradle
raphw: This is a bit naturally grown and should certainly be fixed.

raphw (approved):
Looks good to me, too. Clean and well-documented.

PR #588, opened 2026-03-15
Set Java support version to 8 in root build.gradle

hanggrian:
> Thank you for submitting a pull request! But first:
> 
>  - [ ] Can you back your code up with tests?
> 
>   No tests were added because there is no change in the source code, only Gradle scripts.
> 
>  - [x] Please run `./gradlew spotlessApply :tests:spotlessApply` for auto-formatting.

JVM toolchain is set to 11 since version 6.2.1. JVM toolchain locks the installation to the correct version in the local system, which is a good practice. However, without specifying the target, the compiled bytecode also requires 11:

```
Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Specify proper '-jvm-target' option.
```

A large portion of Java projects still target 1.8.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,5 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
 plugins {
     id "org.gradle.base"
     id "org.shipkit.shipkit-changelog" version "1.2.0"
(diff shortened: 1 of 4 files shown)


jselbo:
Well the historical context here is originally https://github.com/mockito/mockito-kotlin/commit/b1d5e79fd7c78127b43a4b63714c772a33dd4097 

Mockito 5.X originally shipped dropping support for Java 8 runtime, though this was partially walked back due to Android: [#2890](https://github.com/mockito/mockito/issues/2890). As far as I know, there is nothing in mockito-core strictly requiring Java 11 bytecode. So theoretically this would be fine.

Although mockito-kotlin has been shipping with Java 11 bytecode for a while.

@hanggrian could you instead update your tests to build with jvm 11 target?

hanggrian:
Hello, @jselbo.

I was misinformed. I can see now in Mockito's README that only one major version is supported at a time, so lowering the target requirement is against this principle.

I am closing this issue. The hotfix is just what you said, configuring the test tasks to target 11.

```kotlin
tasks {
    compileTestJava {
        options.release.set(11)
    }
    compileTestKotlin {
        compilerOptions.jvmTarget.set(JvmTarget.JVM_11)
    }
}
```

Thanks for taking the time to review my request.

PR #590, opened 2026-04-23
Add benchmarks comparing mockito-kotlin and MockK

jselbo:
The goal here is to compare the runtime performance of Mockito-Kotlin/Mockito with MockK for common mocking workflows. I set up three test cases with equivalent Mockito-Kotlin and MockK code using the latest versions of each library.

It shows that Mockito is 10-40% faster on a cold iteration and 2-3x faster avg. on warm iterations than MockK for the "basic" and "heavy" workflows. 

The "objectMocking" workflow shows a huge difference which seems to be caused by the `unmockkObject` call which re-instruments the class. In contrast, Mockito does not re-instrument the class when closing the `MockedObject` handle (although it no longer intercepts invocations).

I'm not pretending to be an expert in benchmarking, so let me know if there is a way to make this more robust.


hick209 (approved):
Wow! Very interesting results here. I had heard that Mockito was faster, but I was surprised by how much on the more complex cases there.

raphw:
That sounds intuitive with Mockito aiming to avoid reinstrumentation as much as it can be done. Good you created a showcase.

PR #591, opened 2026-04-23
Update README

jselbo:
- Fix badge. `maven-badges.herokuapp.com` is dead and `oss.sonatype.org` is deprecated and doesn't show latest version.
- Add Kotlin Gradle build file example
- Link to Android wiki since I think this is a prominent use case for Mockito-Kotlin

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,22 @@
 # Mockito-Kotlin
 
-[ ![Download](https://maven-badges.herokuapp.com/maven-central/org.mockito.kotlin/mockito-kotlin/badge.svg) ](https://maven-badges.herokuapp.com/maven-central/org.mockito.kotlin/mockito-kotlin)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,19 +1,24 @@
 # Mockito-Kotlin
 
-[ ![Download](https://maven-badges.herokuapp.com/maven-central/org.mockito.kotlin/mockito-kotlin/badge.svg) ](https://maven-badges.herokuapp.com/maven-central/org.mockito.kotlin/mockito-kotlin)
```
hick209, line 3: Why remove the badge? I see it may not be working currently, but maybe it's fixable or we could replace with a different one, WDYT?

I have some setup in ktfmt, so feel free to copy it. There are also of examples online of course.
https://github.com/facebook/ktfmt/

hick209 (approved):
LGTM, just one minor comment

README.md
jselbo: I think we just need the one Maven badge. I'm not sure why there were two separate ones before.

PR #592, opened 2026-05-18
Add InOrder verification to MockedStatic

mirkoalicastro:
Adding the capability to verify static method calls in order.
Also, forwarding InOrder's default method explicitly, since Kotlin class delegation does not generate forwarders for Java default methods.

 - [x] Please run `./gradlew spotlessApply` for auto-formatting.

### DIFF
--- a/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/KInOrderDecorator.kt
+++ b/mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/KInOrderDecorator.kt
@@ -26,10 +26,29 @@
 package org.mockito.kotlin.internal
 
 import org.mockito.InOrder
+import org.mockito.MockedStatic
 import org.mockito.kotlin.KInOrder
 import org.mockito.verification.VerificationMode
 
 class KInOrderDecorator(private val inOrder: InOrder) : KInOrder, InOrder by inOrder {
+    /**
+     * Verifies static interaction in order, with exactly one number of invocations.
+     *
+     * Forwarded explicitly because Kotlin class delegation (`InOrder by inOrder`) does not generate
+     * forwarders for Java default methods.
+     */
+    override fun verify(mock: MockedStatic<*>, verification: MockedStatic.Verification) {
+        inOrder.verify(mock, verification)
+    }
+
+    override fun <T> verify(
+        mock: MockedStatic<T>,
+        mode: VerificationMode,
+        verification: MockedStatic.Verification,
+    ) {
+        inOrder.verify(mock, verification, mode)
+    }
+
     override fun <T> verifyBlocking(mock: T, f: suspend T.() -> Unit) {
         val m = verify(mock)
         safeRunBlocking { m.f() }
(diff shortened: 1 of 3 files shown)


tests/src/test/kotlin/test/MockedStaticTest.kt
```diff
@@ -16,4 +18,52 @@ class MockedStaticTest : TestBase() {
             mocked.verify(times(2)) { SomeObject.aStaticMethod() }
         }
     }
+
+    @Test
+    fun testInOrderVerifyStatic() {
+        mockStatic(SomeObject::class.java).use { mocked ->
```
jselbo: Please use the reified `mockStatic<T>` in these tests

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/KInOrderDecorator.kt
```diff
@@ -26,10 +26,29 @@
...
 import org.mockito.verification.VerificationMode
 
 class KInOrderDecorator(private val inOrder: InOrder) : KInOrder, InOrder by inOrder {
+    /**
+     * Verifies static interaction in order, with exactly one number of invocations.
+     *
+     * Forwarded explicitly because Kotlin class delegation (`InOrder by inOrder`) does not generate
+     * forwarders for Java default methods.
+     */
+    override fun verify(mock: MockedStatic<*>, verification: MockedStatic.Verification) {
```
jselbo: In my testing this override doesn't seem necessary. This default method already invokes `verify(MockedStatic<?>, MockedStatic.Verification, VerificationMode)` signature (a non-default method) which is correctly forwarded. Am I missing something?

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/KInOrderDecorator.kt
```diff
@@ -26,10 +26,29 @@
...
+        inOrder.verify(mock, verification)
+    }
+
+    override fun <T> verify(
```
jselbo, line 34: This one still seems like a worthwhile addition to mirror the `MockedStatic.verify(VerificationMode, MockedStatic.Verification)` extension fun we have so callers can put verification lambda last. ~~I would add a comment like the one on that extension fun~~

edit: I see the comment on the interface now. Looks good

mirkoalicastro:
Hi @jselbo . Any chance you can review this PR? 🙂

jselbo:
@mirkoalicastro  So sorry, I had written comments but didn't click the submit button.

tests/src/test/kotlin/test/MockedStaticTest.kt
mirkoalicastro: done, thanks

mockito-kotlin/src/main/kotlin/org/mockito/kotlin/internal/KInOrderDecorator.kt
mirkoalicastro: You're right. I added it to silence the intellij inspection:
> Java default methods are not overridden by delegation

But since InOrderImpl does not override the default, the inspection is a false positive. I removed it for now, although keeping it would keep the class free from intellij inspection warnings. I'll leave this to your decision, if you prefer to re-add it, I'm happy to do it.

PR #593, opened 2026-05-22
Introduce spyObject

jselbo:
This seems useful to have - same as other spy APIs, this configures the mocks with `CALLS_REAL_METHODS` answer.

_____________________

 - [X] Please run `./gradlew spotlessApply` for auto-formatting.


hick209 (approved):
Simple enough and looks like a great addition to the current API.
Awesome work! 👏

